PR #49, opened 2013-04-11
Enable fingerprint for non-module Prop

som-snytt:
`Framework.tests` appears to have a typo that disables detection of tests
that aren't modules.  Maybe that's intentional.

Probably every test in existence is a module extending Properties.

The ! line doesn't include the class name as property name. And the count is 2 instead of 1.

From sbt, `No main class detected.`

I'm exercising test-interface from partest, but this is not a need-driven PR.


som-snytt:
https://github.com/typelevel/scalacheck/issues/92

PR #93, opened 2014-02-28
Added features for  more advanced collections of data and examples

edmondo1984:
Hello, 
I added some very simple code to explain how to collect stats on the input data. I think this should very useful and I added it to my wiki fork as well


rickynils:
I'm going to close this because I think this is something that shouldn't be part of core ScalaCheck. I want to keep ScalaCheck minimal, and while we need improvement in this area, I need to think about what to to with the current command line runner. I'd like to remove it entirely, and just keep the sbt runner. Then we could have a separate project for a ScalaCheck-specific test runner.

PR #133, opened 2015-01-15
add Arbitrary#fromFunction. more convenient way to define Arbitrary instances

xuwei-k:
example

```scala
case class Foo(a: Int, b: String)

// before
implicit val fooArbitrary = Arbitrary(for {
  a <- implicitly[Arbitrary[Int]].arbitrary
  b <- implicitly[Arbitrary[String]].arbitrary
} yield Foo(a, b))

// after
implicit val fooArbitrary =
  Arbitrary.fromFunction(Foo.apply _)
```


softprops:
+1

PR #135, opened 2015-01-15
Implemented a generic Numeric Shrinker

DomBlack:
This uses the same logic as the original Int Shrinker, but will cover all Integral
and Fractional data types in Scala, including ones missing generators such as
BigDecimal.

Related Issue: rickynils/scalacheck#130

**Note; this is only an initial pass, but I've run out of time to improve it.**


rickynils:
I've merged this in 93f49193276c859a16b7d13ad24dc51faabf9a30. Also did some fixes in 8843a44aacac6fb785f59fcabfba14430a068aa2.

Thanks!

Issue #122, opened 2014-12-02
Port to Scala.js 0.6 and its new test interface


InTheNow:
Similar to [Port uTest to Scala.js 0.6](https://github.com/lihaoyi/utest/issues/44).

Release details are [on the Scala.Js forum](https://groups.google.com/forum/?fromgroups#!topic/scala-js/QMlQwSt4VS0)

bblfish:
+1 for me on this.

antonkulaga:
+1 for me on this.

I also want to point out that scalajs0.6 cross project can be used here to share sources

```
lazy val scalacheck = CrossProject.apply("scalacheck",new File("."),CrossType.Full).
  settings(sharedSettings: _*).
  jsSettings(
    libraryDependencies += "org.scala-js" %% "scalajs-test-interface" % scalaJSVersion
  ).
  jvmSettings(
    libraryDependencies ++= Seq(
      "org.scala-sbt" % "test-interface" % "1.0",
      "org.scala-js" %% "scalajs-stubs" % scalaJSVersion % "provided"
    )
  )

lazy val js = scalacheck.js
lazy val jvm  = scalacheck.jvm
```

InTheNow:
Quick heads up that I've done an initial port to scalajs 0.6. It's working OK, but needs polishing before I can push for initial feedback, hope in next day or two. This is using RC1, so looks like we should be ready for the final 0.6 release ( that is a BIG one, btw).

In meantime, chats are on https://gitter.im/scala-js/scala-js and https://gitter.im/w3c/banana-rdf

InTheNow:
Please see pull request #139, already reviewed by @srjd  and new commit based on review comments.

@srjd - many thanks (again) for your comprehensive review comments.

PR #139, opened 2015-01-25
Feature/scalajs 0.6

InTheNow:
Moved to latest sbt test interface to provide scalajs compatibility (hence number of changes)


build.sbt
```diff
@@ -62,14 +62,17 @@ lazy val sharedSettings = mimaDefaultSettings ++ Seq(
 
 lazy val js = project.in(file("js"))
   .settings(sharedSettings: _*)
-  .settings(scalaJSSettings: _*)
   .settings(
-    libraryDependencies += "org.scala-lang.modules.scalajs" %% "scalajs-test-bridge" % scalaJSVersion,
-    ScalaJSKeys.scalaJSTestFramework in Test := "org.scalacheck.ScalaCheckFramework"
-  )
+    scalaJSStage in Global := FastOptStage,
```
sjrd: This is a personal preference that should typically not make it into the repo.

build.sbt
```diff
@@ -62,14 +62,17 @@ lazy val sharedSettings = mimaDefaultSettings ++ Seq(
...
+    parallelExecution in Test := false,
```
sjrd: Any particular reason for this setting?

build.sbt
```diff
@@ -62,14 +62,17 @@ lazy val sharedSettings = mimaDefaultSettings ++ Seq(
...
+    fork := false,
```
sjrd: What is this supposed to do?

build.sbt
```diff
@@ -62,14 +62,17 @@ lazy val sharedSettings = mimaDefaultSettings ++ Seq(
...
+    testFrameworks := Seq(new TestFramework("org.scalacheck.ScalaCheckFramework"))
+  ).enablePlugins(ScalaJSPlugin)
 
 lazy val jvm = project.in(file("jvm"))
   .settings(sharedSettings: _*)
   .settings(
-    libraryDependencies += "org.scala-sbt" %  "test-interface" % "1.0"
-  )
+    libraryDependencies += "org.scala-sbt" %  "test-interface" % "1.0",
+    libraryDependencies += "org.scala-js" %% "scalajs-stubs" % scalaJSVersion
```
sjrd: You probably want to put this in `% "provided"` so that it does not pollute the publicly visible dependencies in the JVM artifact. Note that this project contains only macros in the JVM version.

src/main/scala/org/scalacheck/MasterRunner.scala
```diff
@@ -0,0 +1,67 @@
...
+    val done = results.doneCount.get
+    val test = results.testCount.get
+    val success = results.successCount.get
+    val failed = results.failureCount.get
+    val errors = results.errorCount.get
+
+    if (slaves > 0) {
+      throw new IllegalStateException(
+        s"There are still $slaves slaves running")
+    }
```
sjrd: I gather you have this code (and the `registered != done` below) because you mimicked the testing framework made for testing testing (double intended) frameworks in Scala.js, aka https://github.com/scala-js/scala-js/blob/master/sbt-plugin-test/testFramework/src/main/scala/sbttest/framework/MasterRunner.scala

If yes, then you should simply remove all of this. We have this in our dummy testing frameworks to test our implementation of the sbt test interface, but in real testing frameworks, this is useless.

Which means you can get rid of `slaveCount`, `registeredCount`, `doneCount`, and the messages that maintain these.

sjrd: Actually, maybe you don't even need a `MasterRunner` and a `SlaveRunner`. [MiniTest does it all with one single `Runner` class](https://github.com/monifu/minitest/tree/master/shared/main/scala/minitest/runner)

src/main/scala/org/scalacheck/MasterRunner.scala
```diff
@@ -0,0 +1,67 @@
...
+      slaveCount.incrementAndGet()
+      // Send Hello message back
+      Some("Hello")
+    case 't' =>
+      // Slave notifies us of registration of tasks
+      registeredCount.addAndGet(msg.tail.toInt)
+      None
+    case 'd' =>
+      // Slave notifies us of completion of tasks
+      results.add(Results(msg.tail))
```
sjrd: Ah, I think that because you want to do this, you need a master and a slave. But besides that message to rather the `Results`, it seems you can get rid of all the other messages.

src/main/scala/org/scalacheck/SlaveRunner.scala
```diff
@@ -0,0 +1,61 @@
...
+    super.serializeTask(task, serializer)
+  }
+
+  override def deserializeTask(task: String,
+                               deserializer: String => TaskDef): Task = {
+    ensureSeenHello()
+    super.deserializeTask(task, deserializer)
+  }
+
+  private def ensureSeenHello(): Unit = {
```
sjrd: Again, this method is irrelevant in a real testing framework.

build.sbt
InTheNow: removed

build.sbt
InTheNow: removed

build.sbt
InTheNow: removed

build.sbt
InTheNow: Added % "provided"

src/main/scala/org/scalacheck/MasterRunner.scala
InTheNow: removed  slaveCount, registeredCount, doneCount, kept slave runner as per next comment

src/main/scala/org/scalacheck/SlaveRunner.scala
InTheNow: removed ensureSeenHello()

InTheNow:
Updated with review feedback.

@sjrd Many thanks for your time and valuable feeback :)

betehess:
Does anybody have an idea for when this has a chance to get merged and released? There are quite a number of tools/libs that depend on scala.js 0.6.x now :-/

rickynils:
@InTheKnow This is a rather intrusive PR, and I would need to set aside some time to review it. One thing that would help, is if the commits are cleaned up a bit. There are things added by one commit and removed by another. Please squash those things, and (if possible) try to separate the commits logically.

I can see that you have spent a good amount of time on this and I really appreciate your effort. I'll try my best to find time to merge this.

InTheNow:
Would it perhaps be easier if I leave this PR as is and do a new PR with just the one, final commit? There isn't really any incremental changes, just one big one to the latest sbt test interface

rickynils:
You can force-push to this PR instead of creating a new one.

InTheNow:
The travis job has failed but the file it was trying to get is actually there - is it possible to restart the test?

rickynils:
@InTheNow, I simplified your PR and pushed 30e9cf2fad568756df4179029499625b7f39d9a9 that closes #122. Thank you for doing the hard work!

I'll probably release 1.12.2 soon to make this accessible to people.

InTheNow:
Thanks!!

PR #141, opened 2015-02-07
command sequence

dpolivaev:
This pull request adds a command sequence defined by Seq[Command]. It makes possible to define a sequence of commands which can be helpful e.g. if some commands usually run in some fixed order.

I shall create a separate pull request where command sequence is extended to support monad like command sequences, which are automatically aborted if any command fails.

Usage example:

```scala
package counter
import org.scalacheck.commands.Commands
import scala.util.Try
import scala.util.Success
import scala.util.Failure
import org.scalacheck.Gen
import org.scalacheck.Prop

case class Counter(private var n: Int = 0) {
  def increment(): Int = {
    if(n == Int.MaxValue)
      throw new IllegalStateException
    else {  
      n = n + 1
      n
    }
  }
  def get: Int = n
}


object CounterProps extends Properties("Counter") {
  property("Counter") = CounterCommandsSpec.property()
}

object CounterCommandsSpec extends Commands {
  import org.scalacheck.Arbitrary.arbitrary
  import org.scalacheck.Gen.oneOf
  type Sut = Counter
  type State = Long
  def newSut(state: State) = new Counter(state.toInt)
  def genInitialState = arbitrary[Int].map(_.toLong)
  def genCommand(state: State) = Gen.const(commandSequense(Increment, Get))
  def canCreateNewSut(newState: State, initSuts: Traversable[State], runningSuts: Traversable[Sut]): Boolean = true
  def destroySut(sut: Sut) {}
  def initialPreCondition(state: State): Boolean = true

  case object Increment extends Command {
    type Result = Int
    def run(counter: Sut) = counter.increment()
    def nextState(state: State) = if (state < Int.MaxValue) state + 1 else state
    def preCondition(state: State) = true
    def postCondition(state: State, result: Try[Int]) =
     result.isSuccess  == (state < Int.MaxValue)
  }

  case object Get extends Command {
    type Result = Int
    def run(counter: Sut) = counter.get
    def nextState(state: State) = state
    def preCondition(state: State) = true
    def postCondition(state: State, result: Try[Int]) =
      result == Success(state)
  }

}

```

### DIFF
--- a/src/main/scala/org/scalacheck/commands/Commands.scala
+++ b/src/main/scala/org/scalacheck/commands/Commands.scala
@@ -158,6 +158,36 @@ trait Commands {
     def preCondition(state: State) = true
     def postCondition(state: State, result: Try[Null]) = true
   }
+  
+   /** A command sequence that is executed completely even if some commands fail. */
+  def commandSequense(head: Command, tail: Command*) = new CommandSequense(head +: tail)
+
+  case class CommandSequense(commands: Seq[Command]) extends SuccessCommand {
+    lazy val headCommand: Command = if (commands.isEmpty) NoOp else commands.head
+    lazy val tailCommands: Command = if (commands.size <= 1) NoOp else new CommandSequense(commands.tail)
+    type Result = (Try[headCommand.Result], tailCommands.Result)
+    def run(sut: Sut): Result = {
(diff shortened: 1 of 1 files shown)


rickynils:
Close in favor of #142

PR #142, opened 2015-02-07
Abortable command sequence

dpolivaev:
This command sequence variant is built upon my pull request https://github.com/rickynils/scalacheck/pull/141 . It allows to create sequences which can be partially executed if any command fails.

Such sequences also add a check that the not executed commands would not change the sut state when they run.


rickynils:
I squashed your commits into aedbe7d6adbdce8f5b161c9b0d23b2eb0970d394.

Also, I cleaned up the code a bit in ab21066c6a498313236e63459c0c624a9b43db15. I also removed `abortableCommandSequence`, since I think the semantics of it was a bit complicated. If you think it is a valuable functionality, we can re-add it.

PR #151, opened 2015-03-05
Arbitrary instances for all tuple and function arities

cvogt:
with best wishes from @xdotai :)

(uses 784ef23 from https://github.com/rickynils/scalacheck/pull/149 to fix build)


cvogt:
Added another commit containing zip and resultOf for all arities and a test case using resultOf on a 22 field case class' .tupled method

rickynils:
Thanks, great stuff! Could you remove 784ef23 from the PR (it's in master now), and squash 0ee2995 and 1ff317c?

cvogt:
sure, curiosity, how can I get the code base to compile without 784ef23? I am on OSX. Also 784ef23 seems like a reasonable change to me, doing it on the sbt level seems more reliable than symlinks.

cvogt:
@rickynils done

rickynils:
Thanks!

cvogt:
@rickynils could you publish another snapshot so this becomes available? thx

PR #152, opened 2015-03-07
Add fully automatic tree / class hierarchy generators

cvogt:
more from @xdotai :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,24 @@ lazy val sharedSettings = mimaDefaultSettings ++ Seq(
 
   crossScalaVersions := Seq("2.10.4", "2.11.5"),
 
+  libraryDependencies <+= (scalaVersion)("org.scala-lang" % "scala-reflect" % _),
+
+  unmanagedSourceDirectories in Compile += {
+    if (scalaVersion.value.startsWith("2.11."))
+      (baseDirectory in LocalRootProject).value / "src" / "main" / "scala-2.11"
+    else if (scalaVersion.value.startsWith("2.10."))
+      (baseDirectory in LocalRootProject).value / "src" / "main" / "scala-2.10"
+    else ???
+  },
+
+  unmanagedSourceDirectories in Test ++= {
+    if (scalaVersion.value.startsWith("2.11."))
+      Seq((baseDirectory in LocalRootProject).value / "src" / "test" / "scala-2.11")
+    else if (scalaVersion.value.startsWith("2.10."))
+      Seq()
+    else ???
+  },
+
   previousArtifact := Some("org.scalacheck" % "scalacheck_2.11" % "1.12.1"),
 
   unmanagedSourceDirectories in Compile += (baseDirectory in LocalRootProject).value / "src" / "main" / "scala",
--- a/src/main/scala-2.10/org/scalacheck/GenTree.scala
+++ b/src/main/scala-2.10/org/scalacheck/GenTree.scala
@@ -1,0 +1,3 @@
+package org.scalacheck
+/** no content for Scala 2.10 */
+trait GenTree
--- a/src/main/scala-2.11/org/scalacheck/GenMacros.scala
+++ b/src/main/scala-2.11/org/scalacheck/GenMacros.scala
@@ -1,0 +1,77 @@
+package org.scalacheck
+class GenMacroImpls(val c: reflect.macros.blackbox.Context){
+  import c.universe._
+
+  def partialTree[T: c.WeakTypeTag]: Tree = treeHelper[T](true)
+
+  def tree[T: c.WeakTypeTag]: Tree = treeHelper[T](false)
+
+  /** Generates a list of all known classes and traits in an inheritance tree. Does not include subclasses of non-sealed ones. */
+  private def linerizedInheritanceTree(symbol:Symbol): Seq[Symbol] = {
+    def children = symbol.asClass.knownDirectSubclasses.flatMap(linerizedInheritanceTree)
+    symbol match{
+      case s if s.isModuleClass => Seq(s)
+      case s if s.isClass => //  && s.asClass.isCaseClass
+        Seq(s) ++ children
+      case s =>
+        c.abort(c.enclosingPosition, s"Can create generator for $s.")
+    }
+  }      
+
+  private def caseObject(sym: Symbol) =
+    sym.asInstanceOf[reflect.internal.Symbols#Symbol].sourceModule.asInstanceOf[Symbol]
+
+  private val scalacheck = q"_root_.org.scalacheck"
+  private val Gen       =  q"$scalacheck.Gen"
+  private val Arbitrary =  q"$scalacheck.Arbitrary.apply"
+  private def Arbitrary(T: Type) = tq"$scalacheck.Arbitrary[$T]"
+
+  private def treeHelper[T: c.WeakTypeTag](allowOpenHierarchies: Boolean): Tree = {
+      val T = weakTypeOf[T]
+      val symbol = T.typeSymbol
+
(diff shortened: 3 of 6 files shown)


rickynils:
@cvogt How does this compare to scalacheck-shapeless?

cvogt:
I wasn't aware of scalacheck-shapeless. I really like its cleanlyness. @alexarchambault please correct me if this is wrong: As far as I can tell from the tests, scalacheck-shapeless does not allow recursive structures, such as trees. See https://github.com/alexarchambault/scalacheck-shapeless/blob/master/src/test/scala/org/scalacheck/Tests.scala
The explanation sounds like the compiler chokes on too large implicit inference chains, which can be arbitrarily large for recursive structures. My macro-based solution works well for recursive structures.

Also my macro supports creating arbitrary classes, not just case classes. Not sure what Shapeless supports. Maybe @milessabin has an opinion on this topic as well.

milessabin:
scalacheck-shapeless uses shapeless's typeclass deriving infrastructure which supports a wide variety of recursion patterns. See for example [here](https://github.com/milessabin/shapeless/blob/master/examples/src/main/scala/shapeless/examples/derivation.scala).

cvogt:
@milessabin I interpreted @alexarchambault's comment as Shapeless principally supporting these recursion patterns, but the depths created by random generation overwhelming scalac. https://github.com/alexarchambault/scalacheck-shapeless/blob/master/src/test/scala/org/scalacheck/Tests.scala#L62
I didn't look at it in detail.

milessabin:
I think you should look at the detail. That type has no finite unfolding other than by terminating with a null or tying a recursive knot somehow.

shapeless has infrastructure for tying recursive knots, but there's no general way that could be applied automatically in all particular cases. shapeless's deriving mechanism also allows manually written implicit instances to be incorporated into a derivation, so someone trying to derive an instance for such a type would be able to manually provide the minimum needed to close a loop and leave the derivation process to handle everything else.

shapeless also handles Symbol prefixes correctly, which from a quick eyeball your macro doesn't. I think you'll be able to trip it up fairly easily with fairly trivial dependent types.

alexarchambault:
@cvogt As @milessabin said, scalacheck-shapeless is using the typeclass facilities from shapeless, which support a wide range of recursion patterns. But it's still possible to stumble upon cases that fail for unclear reasons. (Same thing with argonaut-shapeless, I'm still using the macro-based codec derivation from argonaut as a fallback from time to time.)

I'm not strongly opiniated about all that. The derivation from shapeless is quite general, so there will still be cases it will support that a macro-based approach won't. Inversely, you may handle better the rare cases in which it fails. Imo, it's interesting to have both, and they are the fallback of each other!

cvogt:
I looked at it again and realize I misread the test code regarding recursion. If scalacheck-shapeless does everything and more/better, I am more than happy. Shapeless is the cleaner abstraction for sure. Closing this. Thx guys.

cvogt:
I tried it but didn't have much success with scalacheck-shapeless. In particular migrating my test
* doesn't compile, because there is no implicit Arbitrary available when the tree contains non case-classes.
* Removing the non-case classes makes it compile, but gives a stack overflow (see below), probably because of the recursion.
* Breaking the recursion in the type allowed me to generate objects, but I saw only instances of classes of the first inheritance depth level. I'd however like instances of classes anywhere in the tree with potentially multiple intermediate sealed traits.
* There is no built-in mechanism to safeguard me from forgetting to seal some trait and thus not getting any subclass instances generated. That's why I have `tree`, which fails for this at compile time and `partialTree` which makes it obvious that it may not be complete.

I'd be in favor of a Shapeless-based implementation that satisfies the above use cases. It seems cleaner. But until then I suggest to merge the tree macros, which solve the above uses cases. I am happy to make adjustments beforehand if desired.

```
    ...
	at org.scalacheck.Shapeless$$anonfun$instanceArbitrary$1.apply(Shapeless.scala:42)
	at org.scalacheck.Shapeless$$anonfun$instanceArbitrary$1.apply(Shapeless.scala:42)
	at org.scalacheck.Arbitrary$$anon$3.arbitrary$lzycompute(Arbitrary.scala:66)
	at org.scalacheck.Arbitrary$$anon$3.arbitrary(Arbitrary.scala:66)
	at org.scalacheck.Shapeless$$anonfun$cconsEquallyWeightedArbitrary$1.apply(Shapeless.scala:33)
	at org.scalacheck.Shapeless$$anonfun$cconsEquallyWeightedArbitrary$1.apply(Shapeless.scala:32)
	at org.scalacheck.Arbitrary$$anon$3.arbitrary$lzycompute(Arbitrary.scala:66)
	at org.scalacheck.Arbitrary$$anon$3.arbitrary(Arbitrary.scala:66)
	at org.scalacheck.Shapeless$$anonfun$cconsEquallyWeightedArbitrary$1.apply(Shapeless.scala:34)
	at org.scalacheck.Shapeless$$anonfun$cconsEquallyWeightedArbitrary$1.apply(Shapeless.scala:32)
	at org.scalacheck.Arbitrary$$anon$3.arbitrary$lzycompute(Arbitrary.scala:66)
	at org.scalacheck.Arbitrary$$anon$3.arbitrary(Arbitrary.scala:66)
	at org.scalacheck.Shapeless$$anonfun$instanceArbitrary$1.apply(Shapeless.scala:42)
	at org.scalacheck.Shapeless$$anonfun$instanceArbitrary$1.apply(Shapeless.scala:42)
	at org.scalacheck.Arbitrary$$anon$3.arbitrary$lzycompute(Arbitrary.scala:66)
	at org.scalacheck.Arbitrary$$anon$3.arbitrary(Arbitrary.scala:66)
```

rickynils:
@cvogt I don't think putting this in scalacheck core is a good idea. Since there's obviously different solutions in this problem area, I don't want make one seem more "authorative". Also, I really don't see this as core functionality.

As a side-note, for a long time I've been pondering having some sort of scalacheck-contrib or scalacheck-extra repo where we can be a bit more genereous with dependencies etc. In such a repo we could have both your solution and scalacheck-shapeless, if it would make sense. But, these contrib repos often grow large and unmanageable, so maybe it is just better to have a spread of small repos each implementing some non-core functionality of scalacheck. I don't know what approach is best.

milessabin:
@cvogt your tests ought to work more or less as written with scalacheck-shapeless. If you move the non-recursive cases out of the body of the property then they pretty much do. I'll look into a fix in shapeless to support the nesting correctly.

The form or recursion exhibited by your `Tree` should be fine, but it does SO as you've observed. @alexarchambault could you take a look ... I think that for some reason if a non-`Leaf` is selected then non-leaves will be selected forever after, but I can't immediately see why that should be.

In any case, useful test cases for both shapeless and scalacheck-shapeless ... thanks!

ceedubs:
@rickynils for what it's worth, I would put in a strong vote for a collection of smaller repos as opposed to one scalacheck-contrib repo. I love the shapeless-contrib project, but I've hardly ever been able to use it because it's been rare that its been compatible versions of all of scala, scalaz, shapeless, and scalacheck with the code base I've wanted to use it in. But shapeless-contrib is forced into only releasing with very specific versions of these projects since they all have to move in step for a release.

Sorry to kind of hijack this thread for a scalacheck tangent :(

cvogt:
@rickynils Sounds good. I'll release it separately. Posting a reference here once I did.
@milessabin @alexarchambault Glad this produced some insightful test case.

cvogt:
@rickynils In case you didn't do it yet, could you release another snapshot so I can depend on that? the tree macro requires my changes you recently merged.

cvogt:
released: https://github.com/cvogt/scalacheck-extensions

PR #156, opened 2015-04-12
Bug fix for https://github.com/rickynils/scalacheck/issues/138

dpolivaev:
In the multi threaded test destroySut was called before the property check was executed

### DIFF
--- a/src/main/scala/org/scalacheck/commands/Commands.scala
+++ b/src/main/scala/org/scalacheck/commands/Commands.scala
@@ -198,17 +198,24 @@ trait Commands {
         sutId match {
           case Some(id) =>
             val sut = newSut(as.s)
+            def removeSut  {
+              suts.synchronized {
+                suts -= id
+                destroySut(sut)
+              }
+            }
             val doRun = suts.synchronized {
               if (suts.contains(id)) {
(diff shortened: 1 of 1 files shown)


rickynils:
Great! Thanks!

src/main/scala/org/scalacheck/commands/Commands.scala
```diff
@@ -293,15 +300,21 @@ trait Commands {
   }
 
   /** A property that runs the given actions in the given SUT */
-  private def runActions(sut: Sut, as: Actions): Prop = {
+  private def runActions(sut: Sut, as: Actions, finalize : =>Unit): Prop = {
+    try{
     val (p1, s, rs1) = runSeqCmds(sut, as.s, as.seqCmds)
     val l1 = s"initialstate = ${as.s}\nseqcmds = ${prettyCmdsRes(as.seqCmds zip rs1)}"
     if(as.parCmds.isEmpty) p1 :| l1
     else propAnd(p1 :| l1, {
```
dpolivaev: I am afraid I have not considered a case where property p1 fails here. I should add a propOr, I shall check it and submit a follow-up pull request if needed.

PR #161, opened 2015-05-01
Added `delay` and `protect` for delaying execution of properties.

runarorama:
This addresses a problem I run into frequently with `Properties` being a mutable map.

``` scala
object Test extends Properties("foo") {
  property("bar") = secure {
    val c = connectToTheNetwork()
    val x = something(c)
    c.close()
    p(x)
  }
}
```

In the above, `connectToTheNetwork()` runs _at object creation time_ (!!!). This can cause all kinds of strange initialization errors and (more frequently) blocked threads. With this PR, you would instead say:

``` scala
object Test extends Properties("foo") {
  property("bar") = protect {
    val c = connectToTheNetwork
    val x = something(c)
    c.close()
    p(x)
  }
}
```

and nothing runs at object creation time, other than the creation of the `Prop` which is completely lazy.

### DIFF
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -452,9 +452,23 @@ object Prop {
   def classify(c: => Boolean, ifTrue: Any, ifFalse: Any)(prop: Prop): Prop =
     if(c) collect(ifTrue)(prop) else collect(ifFalse)(prop)
 
-  /** Wraps and protects a property */
+  /**
+   * Wraps and protects a property, turning exceptions thrown
+   * by the property into test failures.
+   */
   def secure[P <% Prop](p: => P): Prop =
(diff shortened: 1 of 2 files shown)


rickynils:
Nice! I'm wondering if we should synchronize the naming with Gen.lazy and Gen.wrap somehow (possibly renaming the Gen functions, wrap is a pretty bad name).

Also, maybe Properties.property should simply be changed to always delay things (I haven't thought of the implications of such a change).

runarorama:
Those are good ideas. I think ultimately I want ScalaCheck to be able to support monadic properties, as per http://hackage.haskell.org/package/QuickCheck-2.1.0.2/docs/Test-QuickCheck-Monadic.html

rickynils:
Yes, monadic properties would be nice. Maybe in ScalaCheck 2... I'm merging this PR, I think your naming is fine, I will adapt `Gen` after it in the future.

PR #163, opened 2015-05-13
Run the tests of the JS platform using Node.js instead of Rhino.

sjrd:
This fixes the Travis builds, because Node.js is just so much faster and less memory-hungry.

After this PR, you may even want to consider bumping up the number of TESTS to be run in the JS env to something like 100 instead of just 20, because it's so much faster now.


rickynils:
Thanks!

PR #171, opened 2015-06-12
Support non-constant arbitrary functions

### LINKED ISSUE #136: Improve arbitrary function generation
This has been discussed for some time. The current way of generating functions are quite stupid, since it will only generate constant functions.

#66 does some changes in this area, but it must be updated to current master. And maybe we should take the Haskell QuickCheck route instead, using some fashion of CoArbitrary.

non:
This PR adds support for better `Arbitrary[FunctionN[...]]` instances. It does a few things:

1. Removes most use of `scala.util.Random`.
2. Implement a simple linear-congruent RNG (Knuth's MMIX), that uses 64-bits of state.
3. Change `Gen` methods to explicitly receive (and pass) RNG seeds (`Long` values)
4. Introduce `Cogen` type to permutate RNG seed states.
5. Introduce a bunch of implicit `Cogen` instances for many built-in types.
6. Use codegen to wire up new `Gen` and `Arbitrary` instances for functions
7. Add a line to `build.sbt` to get tests running faster for ScalaJS.

There is a bit of sketchiness in `Gen` due to its use of `Option[A]` instead of `A`. I was forced to use `.get` (like a few other existing methods do) which is not ideal. I can't tell if you'd expect this to be a major problem for ScalaCheck users.

We should also add a bunch more `Cogen` instances. I chose not to do this until I got a sense of whether this PR was going in the right direction.

Finally, while I bet most people using arbitrary functions would use primitive types, there is a bunch of work around updating documentation to explain this approach (and how to define custom `Cogen` instances).

Fixes #136


non:
Argh! Tests passed locally -- will investigate why they are failing for Travis.

non:
Oh, OK. These are all MIMA failures.

This change is definitely not binary-compatible.

xuwei-k:
awesome! But maybe there is a same issue as my library.

https://github.com/scalaprops/scalaprops/issues/6

```
scala> org.scalacheck.Gen.listOfN(10, implicitly[Arbitrary[Int => Int]].arbitrary).sample.get.map((1 to 10).map(_)) foreach println
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
Vector(-2147483648, 1, -2147483648, 205452885, 2043426884, 1, -1632521114, 1, -2147483648, 243620675)
```

non:
Ouch! Good catch. Yes, that's definitely a bug.

src/main/scala/org/scalacheck/Cogen.scala
```diff
@@ -0,0 +1,103 @@
...
+ **  This software is released under the terms of the Revised BSD License.  **
+ **  There is NO WARRANTY. See the file LICENSE for the full text.          **
+ \*------------------------------------------------------------------------ */
+
+package org.scalacheck
+
+import language.higherKinds
+import language.implicitConversions
+
+sealed trait Cogen[-T] {
```
xuwei-k: `sealed abstract class` is better for keep binary compatibility.

xuwei-k:
following expression should returns `List((false,false), (false,true), (true,false), (true,true))`

```
scala> org.scalacheck.Gen.listOfN(100000, implicitly[Arbitrary[Boolean => Boolean]].arbitrary).sample.get.map{f => (f(true), f(false))}.distinct.sorted
res0: List[(Boolean, Boolean)] = List((false,true), (true,false))

scala> org.scalacheck.Gen.listOfN(100000, implicitly[Arbitrary[Boolean => Boolean]].arbitrary).sample.get.map{f => (f(true), f(false))}.distinct.sorted
res1: List[(Boolean, Boolean)] = List((false,true), (true,false))

scala> org.scalacheck.Gen.listOfN(100000, implicitly[Arbitrary[Boolean => Boolean]].arbitrary).sample.get.map{f => (f(true), f(false))}.distinct.sorted
res2: List[(Boolean, Boolean)] = List((false,true), (true,false))
```

rickynils:
@non This is great stuff! I have worked in a similar direction in the `scalacheck2` branch, but there I've made even more changes to `Gen` to also support finite generators. Due to lack of time, lots of stuff remains to be done there, though.

So the question is how to go forward. The realistic approach is probably to try to get the immutable gen and cogen into a 1.13 release, and keep my other changes for 2.0.

I'll make an attempt to dig into your solution as soon as possible.

non:
There does seem to be a problem with `Gen.oneOf` combined with functions, which is demonstrated by @xuwei-k's second issue. I'm looking into it.

@rickynils that sounds fine. Let me know if you see any issues or want me to make any changes.

rickynils:
@non If you rebase this PR, you can bump `versionNumber` in `build.sbt` to `1.13.0`, and set `previousArtifact` to `None`. That should take care of the failing Travis tests.

non:
Alright, I just did what you suggested: I rebased against `upstream/master` and edited `build.sbt`.

I'm still working on the issue that @xuwei-k noticed with `Arbitrary[Boolean => Boolean`. I'll push a commit and comment here once I have that issue fixed.

non:
These last commits fix the issue that @xuwei-k found. At this point I consider this PR feature complete.

@rickynils is there anything else you'd like me to do? Scalacheck probably needs documentation explaining how `Cogen[A]` works but I wasn't sure if you wanted me to tackle that in this PR or not.

xuwei-k:
:+1:

jedesah:
This is awesome!

rickynils:
Now that snapshots are published automatically, you can try out this in `1.13.0-ea2bbe6-SNAPSHOT`

lilred:
I found this page while looking for documentation on the topic of `Cogen`. I don't think there is such a thing at the moment, but you can check [this Haskell blog post](https://kseo.github.io/posts/2016-12-14-how-quick-check-generate-random-functions.html) to get the gist of it.

PR #178, opened 2015-07-10
add colors to console output

ChrisCoffey:
This adds green/red coloring to log output (where supported), making visual test inspection much simpler.


xuwei-k:
-1

https://github.com/sbt/test-interface/blob/v1.0/src/main/java/org/scalatools/testing/Logger.java#L8

should respect `Logger#ansiCodesSupported`

ChrisCoffey:
Good call, I'll clean that up.

rickynils:
Thanks!

rickynils:
I'll merge it when I get back to a computer...

PR #179, opened 2015-07-11
Use Travis Build Containers

ChrisCoffey:
Some of the scalacheck tests fail with code 137, which usually indicates they're running up against Travis' resource limits. The [new Travis infrastructure] (http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade) provides "...2 dedicated cores and 4GB of memory, vs 1.5 cores and 3GB...", which should alleviate the issue.


rickynils:
Great stuff! Thank you!

Issue #187, opened 2015-08-04
Custom number of runs per property


sid-kap:
I'm using scalacheck with sbt. When I run the `test` command in sbt, Scalacheck, by default, generates 100 test cases for each `Prop` in all `Properties` classes in my test path.

Can we add a feature so that I can specify minSuccessfulTests for each Prop in my source code? This would be useful so that I can run fewer tests for tests that take longer to run.

What I have in mind is an annotation that I can apply to either my Properties class or each property block in my Properties class, specifying minSuccessfulTests for the appropriate tests.

I think this would be easy to implement, so I wouldn't mind implementing this if you guys like the idea.

rickynils:
This should be possible to add in ScalaCheck 2. Adding it to the current implementation would probably be too involved.

sid-kap:
I'm not sure that I want to wait until ScalaCheck 2.
I think I'll take a stab at this and see if we can get this working without changing the current implementation too much?

rickynils:
PR #188 has been merged.

PR #188, opened 2015-08-05
Allow overriding of Test.Parameters in Properties

sid-kap:
This is a basic implementation of the feature request in #187.
This would allow a user to do something like 

```scala
object MyProperties extends Properties("My object") {
  override def overrideParams(p: Test.Parameters) = p.withMinSuccessfulTests(20)
}
```
if they want Scalacheck to, for this class only, run 20 tests per property instead of whatever value was specified in the sbt argument or the project default.

### DIFF
--- a/src/main/scala/org/scalacheck/Properties.scala
+++ b/src/main/scala/org/scalacheck/Properties.scala
@@ -32,6 +32,12 @@ class Properties(val name: String) {
 
   private val props = new scala.collection.mutable.ListBuffer[(String,Prop)]
 
+  /**
+   * Changes to the test parameters that are specific to this class.
+   * Can be used to set custom parameter values for this test.
+   */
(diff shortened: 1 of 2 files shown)


sid-kap:
How do you feel about this approach?
Would you prefer doing this using an annotation instead?

rickynils:
As you said, this approach works only for property collections. But that is fine for a start, I think. I try to merge this next week.

(for ScalaCheck 2 I have a more general approach going, which should allow for per-property params)

You can also check out specs2, there is a way to set params through implicits there, but I haven't looked at it in detail.

sid-kap:
Hi, have you gotten a chance to look at this yet?

sid-kap:
Hi, have you gotten a chance to look at this yet?

rickynils:
@sid-kap Sorry for my late response... and thanks!

PR #196, opened 2015-10-12
Issue#186 : Fix Gen.Choose(Double.MinValue, Double.MaxValue).

jahangirmohammed:


src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -258,10 +259,17 @@ object Gen extends GenArities{
 
     private def chDbl(l: Double, h: Double)(p: P, seed: Seed): R[Double] = {
       val d = h-l
-      if (d < 0 || d > Double.MaxValue) r(None, seed)
+      val (n, s) = seed.double
+      if (d < 0) r(None, seed)
+      else if (d > Double.MaxValue) {
+        if(Random.nextBoolean()){
```
rickynils: @jahangirmohammed Calling `Random.nextBoolean()` here is not deterministic. Given a `seed` value we want to always return the same result.

jahangirmohammed: @rickynils : thanks for the feedback and time. Am thinking may be we can use re-use `seed.double` value instead of `Random.nextBoolean()` like below:
```
      val (n, s) = seed.double
      if (d < 0) r(None, seed)
      else if (d > Double.MaxValue) {
        if(n > 0.5){
          r(Some((n * -l) + l), s)
        } else {
          r(Some(n * h), s)
        }
      }
      else if (d == 0) r(Some(l), seed)
      else {
        r(Some(n * (h-l) + l), s)
      }
```
Please let me know if this is fine so that I can add this.

rickynils:
This was fixed by #210, so I'm closing this PR

PR #200, opened 2015-10-27
don't require an implicit view for collections to Traversable 

jedws:
don't require an implicit view for collections to Traversable  just to get `forall` and `size`, use typeclasses instead that provide more flexibility and coverage.

The `buildableOf` methods had an implicit view function, but it is better to use the `IsTraversableLike` functionality to get this function, or for Maps just to provide them explicitly.

The `suchThat => forall` guard on `buildableOf` and `buildableOfN` was unnecessary as sequence already does the right thing, this should save a traversal of each built collection, and save a few clock cycles.


rickynils:
Thank you! Much cleaner than the previous solution.

PR #201, opened 2015-10-27
Add flatten on Gen

jedesah:


src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -55,6 +55,9 @@ sealed abstract class Gen[+T] {
     rt.flatMap(t => f(t).doApply(p, rt.seed))
   }
 
+  def flatten[U](implicit asOption: (T) ⇒ Option[U]): Gen[U] =
```
rickynils: Minor thing, but can you use `=>` instead of `⇒` here? I also think you can write just `T` instead of `(T)`.

jedesah:
Done! :smile:

rickynils:
Thanks!

PR #207, opened 2015-11-18
Add the possibility to shrink the state in commands

markus1189:
Adds a `shrinkState` method to the `Commands` trait to allow shrinking of the `State` inside imperative specifications.  I added an override in the `CommandsSpecification` file for the counter example.

Some notes you might want to take into account:

- unfortunately, [this](https://github.com/rickynils/scalacheck/compare/master...markus1189:shrink-state?expand=1#diff-f58118b457fead9d506c771626fa3d5fR233) always resolves to the `anyInstance` for Shrink, which does no shrinking at all, which means you *always* have to overwrite the `shrinkState` method, even if your internal `State` is only `Int` instead of a custom type.  I think we should consider to add this to the user guide somehow.  I would be happy to do this if wanted.

- [here](https://github.com/rickynils/scalacheck/compare/master...markus1189:shrink-state?expand=1#diff-f58118b457fead9d506c771626fa3d5fR252)  I decided to first check for `Actions` that are shrinked for all fields (`state`,`seqCmds`,`parCmds`), is this a sensible order?  I am not sure what the smartest shrinking order is ;)

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/commands/CommandsSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/commands/CommandsSpecification.scala
@@ -28,6 +28,8 @@ object CommandsSpecification extends Properties("Commands") {
     type Sut = Counter
     type State = Int
 
+    override def shrinkState: Shrink[Int] = implicitly
+
     def canCreateNewSut(newState: State, initSuts: Traversable[State],
       runningSuts: Traversable[Sut]) = true
 
--- a/src/main/scala/org/scalacheck/commands/Commands.scala
+++ b/src/main/scala/org/scalacheck/commands/Commands.scala
@@ -244,8 +244,8 @@ trait Commands {
             }
             if (doRun) runActions(sut,as, removeSut)
             else {
-            	removeSut
-            	Prop.undecided
+                removeSut
+                Prop.undecided
             }
 
           case None => // NOT IMPLEMENTED Block until canCreateNewSut is true
@@ -259,6 +259,9 @@ trait Commands {
     }
   }
 
+  /** Override this to provide a custom Shrinker for your internal
+    * [[State]].  By default no shrinking is done for [[State]]. */
+  def shrinkState: Shrink[State] = implicitly
 
   // Private methods //
   private type Commands = List[Command]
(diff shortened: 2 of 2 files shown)


rickynils:
@markus1189 Sorry for the late response! Is it not possible to just define an implicit `Shrink` instance for your `State` type to get this functionality? Like:

```
implicit def shrinkState(s: State): Shrink[State] = ...
```

or am I missing something?

markus1189:
Yes that is what I assumed naively too ;)  Unfortunately, the implicit `anyInstance` (see my bullet point above) is chosen such that there is no state shrinking... :(

rickynils:
Yeah, you're correct. There seems to be no way to get Scala to find an implicit `Shrink` instance within a `Command` subclass. So your solution is probably the best we can get. It's an improvement over the current situation at least.

I have added a small utility method called `Shrink.shrinkWithOrig` that makes the implementation of `shrinkActions` a bit nicer. Can you rebase your PR and then do something like this:

```
  private implicit val shrinkActions: Shrink[Actions] = Shrink[Actions] { as =>
    def shrinkedCmds: Stream[Actions] =
      Shrink.shrink(as.seqCmds).map(cs => as.copy(seqCmds = cs)) append
      Shrink.shrink(as.parCmds).map(cs => as.copy(parCmds = cs))

    Shrink.shrinkWithOrig[State](as.s)(shrinkState) flatMap { state =>
      shrinkedCmds.map(_.copy(s = state))
    }   
  }
```

markus1189:
I adapted the example to use `Shrink.shrinkWithOrig` but I am not sure if I got that correct, maybe you can have another look?

rickynils:
You don't need to do the `append shrinkedCmds`. `shrinkWithOrig` always includes the original (unshrinked) state as the first element in the stream. So this should just work:

```
  private implicit val shrinkActions: Shrink[Actions] = Shrink[Actions] { as =>
    def shrinkedCmds: Stream[Actions] =
      Shrink.shrink(as.seqCmds).map(cs => as.copy(seqCmds = cs)) append
      Shrink.shrink(as.parCmds).map(cs => as.copy(parCmds = cs))

    Shrink.shrinkWithOrig[State](as.s)(shrinkState) flatMap { state =>
      shrinkedCmds.map(_.copy(s = state))
    }   
  }
```

markus1189:
I may be wrong, but I thought that it does make a difference whether we include `shrinkState` again or whether we append `shrinkCmds` at the end.  As I understood the shrinking, it tries every element of the `Stream` until it finds an element for which the property fails.  By only including the unshrinked state, we would not consider the cases where instead of shrinking the `State` we only shrink the `Actions`, leaving the `State` as is?

However I adapted the code to your suggestion ;)

rickynils:
`shrinkWithOrig` gives us a stream with the original state plus the shrinked states: `s, s', s'', s''', ...`. 
`shrinkedCmds` is a stream of shrinked versions of the `Actions` list: `as', as'', as''', ...`.
By flat-mapping those two streams we get all combinations between them. What is missing is just the original `as` with the shrinked states. We could maybe add that case too, for perfection.

rickynils:
Merged in c4cc9ce766ce252b0323d2f796fe62de2f0bab5d

Issue #186, opened 2015-07-27
Gen.choose(double1, double2) fails if (double2 - double1) is infinite


WellingR:
Example `Gen.choose(Double.MinValue, Double.MaxValue)` always fails to generate a double.

I see that in this specific case it would be better to use `Arbitrary.arbitrary[Double]`, however `Gen.chooseNum` has the same problem (it fails unless it generates one of the special values).

PR #210, opened 2015-12-07
Allow Gen.choose to handle a very large range of doubles

suzannehamilton:
A proposed fix for issue #186.

When generating a random double in a range, `chDbl` is calculating max-min, which could potentially be outside of the range of doubles. To prevent this, `chDbl` is just discarding those values, which causes ranges like `Gen.choose(Double.MinValue, Double.MaxValue)` to fail.

As issue #186 says, you're probably better off using `Arbitrary` in this case (they use different generation strategies, and the distribution of doubles produced by `Arbitrary` is probably more useful), but I think it's reasonable to expect `Gen.choose(Double.MinValue, Double.MaxValue)` to generate something.

I've fixed the issue by copying the strategy used in [drmaciver/hypothesis](https://github.com/DRMacIver/hypothesis) for [float generation](https://github.com/DRMacIver/hypothesis/blob/master/src/hypothesis/strategies.py): if max-min is out of the double range, then recursively call the double generator with a range of _either_ min to zero, or zero to max.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -85,8 +85,11 @@ object GenSpecification extends Properties("Gen") {
   }
 
   property("choose-double") = forAll { (l: Double, h: Double) =>
-    if(l > h || h-l > Double.MaxValue) choose(l,h) == fail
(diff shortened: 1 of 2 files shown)


rickynils:
This is great! Thank you!

suzannehamilton:
Thanks!

PR #220, opened 2016-02-11
Added symbolic/dynamic terms to ScalaCheck.Commands

alathon:
Notable points:

1) Unfortunately, due to the way CommandSequence works, I was not able to find any way to 'fix' it with the new arguments added to run/nextState(), so I've had to remove CommandSequence. Mainly, calling nextState() from a preCondition() is a violation of the way QuickCheck works, and cannot be (as far as I can see) represented now that there is a Term argument to nextState().

2) I'm still not entirely convinced on my implementation of Term, but it works without too much effort. CommandPidSpecification shows the process registry example from QuickCheck: For Fun & Profit, which is now possible due to symbolic terms. A more advanced version can be found in CommandMultiPidSpecification, where one method returns a List of PIDs, instead of there being an add method that adds a single one. This is much harder to deal with, because you have no way of knowing how big the List will actually be before run() is executed.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/commands/CommandsSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/commands/CommandsSpecification.scala
@@ -46,24 +46,24 @@ object CommandsSpecification extends Properties("Commands") {
 
     case object Get extends SuccessCommand {
       type Result = Int
-      def run(sut: Sut) = sut.get
-      def nextState(state: State) = state
+      def run(sut: Sut, state: State) = sut.get
+      def nextState(state: State, v:Term[Result]) = state
       def preCondition(state: State) = true
       def postCondition(state: State, result: Result) = result == state
     }
 
     case object Inc extends SuccessCommand {
       type Result = Int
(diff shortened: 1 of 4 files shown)


alathon:
@rickynils Any news on whether you think the implementation of terms proposed here is acceptable, whether you would like more test cases/showcases for usage, etc? I can spend some time considering alternative representations, but only if it seems necessary/appropriate =)

alathon:
I think this needs to go back to the drawing board a bit; I'll test a variety of different implementations against the same set of scenarios / SUTs, and do a new pull request when the time comes.

rickynils:
@alathon I really appreciate your work on this. I'm sorry for my lack of feedback. Due to other commitments, I don't have much time to put in on ScalaCheck right now.

alathon:
@rickynils No problem :) I yanked this pull request primarily because I will be evaluating a few different approaches to symbolic values as part of my thesis, and the work on my fork will hopefully give rise to a decent implementation that you might be able to pull into ScalaCheck later.

PR #222, opened 2016-02-20
exclude '\uFFFE' from Arbitrary[Char] pending jdk bug review

refried:
Although the Unicode spec is clear:

>To ensure that the mapping for a Unicode encoding form is one-to-one, all Unicode scalar values, including those corresponding to noncharacter code points and unassigned code points, must be mapped to unique code unit sequences. Note that this requirement does not extend to high-surrogate and low-surrogate code points, which are excluded by definition from the set of Unicode scalar values.

There is a bug at http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/sun/nio/cs/UnicodeDecoder.java#94 which causes incorrect handling of `'\ufffe'` (which is designated one of those valid "noncharacter code points") in the jdk.  Until they fix the bug by removing that unnecessary test (I submitted a bug report, but it's still pending review and hasn't been assigned a real bug id), I recommend removing `'\ufffe'` from `Arbitrary[Char]`.  Otherwise, basic properties like this will fail:

```scala
object UtfSpec extends Properties("String") {
  property("survives utf16 encode/decode") =
    forAll { (a: String) =>
      new String(a.getBytes("UTF-16"), "UTF-16") == a
    }
}
```
```
! String.survives utf16 encoding: Falsified after 5097 passed tests.
> ARG_0: "?"
> ARG_0_ORIGINAL: "臟묰욇泳帣뭯㌺釶쒛ퟩ▆됻窿襈ꁅ育儝㆙㽦횳싺搜蘠쯏䬝倻樓췝亢跡驯￾賬簘倃両㖕㔹"
```

### DIFF
--- a/src/main/scala/org/scalacheck/Arbitrary.scala
+++ b/src/main/scala/org/scalacheck/Arbitrary.scala
@@ -114,12 +114,18 @@ private[scalacheck] sealed trait ArbitraryLowPriority {
   )
 
   /** Arbitrary instance of Char */
-  implicit lazy val arbChar: Arbitrary[Char] = Arbitrary(
-    Gen.frequency(
-      (0xD800-Char.MinValue, Gen.choose[Char](Char.MinValue,0xD800-1)),
-      (Char.MaxValue-0xDFFF, Gen.choose[Char](0xDFFF+1,Char.MaxValue))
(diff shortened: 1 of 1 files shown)


rickynils:
Nice find! Thank you!

PR #236, opened 2016-04-15
fix mima settings

xuwei-k:
- mima 0.1.6 does not support Scala 2.12.0-M4 https://github.com/typesafehub/migration-manager/issues/78
- fix previous artifact


rickynils:
Nice, thank you!

PR #242, opened 2016-05-31
Replace floatToRawIntBits by floatToIntBits and doubleToRawLongBits by doubleToLongBits

julien-truffaut:
I am not sure it is the best approach, maybe it would be better to use raw methods in jvm world and non raw methods in scala.js. However, I don't know enough about the difference to make the call.


InTheNow:
Some extra notes https://github.com/non/spire/pull/485#issuecomment-126917183

julien-truffaut:
any update?

julien-truffaut:
@rickynils would it be possible to get a new release with this change?

Issue #218, opened 2016-02-05
java.util.NoSuchElementException: None.get


xuwei-k:
https://github.com/scalaz/scalaz/pull/1095

rickynils:
Steps to reproduce:

```
git clone https://github.com/xuwei-k/scalaz.git -b scalacheck-1.13.0
cd scalaz
sbt "testOnly scalaz.CofreeTest -- -v 2"
```

Relevant output:

```
[info] ! scalaz.CofreeTest.CofreeStream:monad.left identity: Exception raised on property evaluation.
[info] > ARG_0: -502843521
[info] > ARG_1: <function1>
[info] > Exception: java.util.NoSuchElementException: None.get
[info] scala.None$.get(Option.scala:313)
[info] scala.None$.get(Option.scala:311)
[info] org.scalacheck.GenArities$$anonfun$function1$1$$anonfun$1.apply(GenArities.scala:15)
[info] scalaz.CofreeBind$class.bind(Cofree.scala:224)
[info] scalaz.CofreeInstances0$$anon$3.bind(Cofree.scala:170)
[info] scalaz.CofreeInstances0$$anon$3.bind(Cofree.scala:170)
[info] scalaz.Monad$MonadLaw$class.leftIdentity(Monad.scala:80)
[info] scalaz.Monad$$anon$3.leftIdentity(Monad.scala:82)
[info] scalaz.scalacheck.ScalazProperties$monad$$anonfun$leftIdentity$4.apply(ScalazProperties.scala:211)
[info] scalaz.scalacheck.ScalazProperties$monad$$anonfun$leftIdentity$4.apply(ScalazProperties.scala:211)
[info] org.scalacheck.Prop$$anonfun$forAll$10$$anonfun$apply$26.apply(Prop.scala:870)
[info] scala.Function1$$anonfun$andThen$1.apply(Function1.scala:55)
[info] org.scalacheck.Prop$$anonfun$forAllShrink$1$$anonfun$3.apply(Prop.scala:712)
[info] org.scalacheck.Prop$$anonfun$forAllShrink$1$$anonfun$3.apply(Prop.scala:712)
[info] org.scalacheck.Prop$.secure(Prop.scala:456)
[info] org.scalacheck.Prop$$anonfun$forAllShrink$1.org$scalacheck$Prop$$anonfun$$result$1(Prop.scala:712)
[info] org.scalacheck.Prop$$anonfun$forAllShrink$1.apply(Prop.scala:749)
[...]

PR #257, opened 2016-08-30
Avoid empty generators where possible.

non:
This commit was inspired by #218. It turns out that it ends up being
very easy to accidentally create empty generators (that will never
return `Some(_)` values) unnecessarily. The particular issue was around
usage of `Gen.sized`, but there are a number of other combinators that
end up creating empty generators in some cases.

Previously, when the arguments to a combinator make it impossible to
produce a value, ScalaCheck has preferred to produce empty `Gen[T]`
values. After this change, combinators whose arguments make it
impossible to generate values will instead throw exceptions during
creation. This makes it easier for the author to realize and correct
their mistake.

The commit also tightens up a bunch of the internal logic around
generating collections, numbers in a given range, etc. to minimize the
use of empty generators.

The reason why empty generators are a much bigger problem now than
before is the introduction of `Cogen[T]`. When we produce an arbitrary
`A => B` value, we do so by using a `Cogen[A]` to permute the PRNG,
and then use a `Gen[B]` to produce a `B` value from that PRNG
state. If our generator is empty (or even if it's a very sparse
partial generator), this process will fail when we try to apply the
function to an `A` value.

Since ScalaCheck allows empty generators (as QuickCheck does) there's
probably no backwards-compatible way to be 100% sure a `Gen[B]` is
safe to use. However, in practice most `Gen` instances *could* be safe
to use this way (as they are in QuickCheck), but are not due to how
easy it is to end up with empty (or mostly empty) generators.

This changes is binary compatible with the last release (check with
MiMA). It does introduce new runtime behavior (in the form of some new
exceptions users might see), but I think the change is worth it to get
faster (and safer) generators, as well as more reliable function
generators.

Review by @rickynils, @xuwei-k, and anyone else interested.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -16,6 +16,7 @@ import Prop.{forAll, someFailing, noneFailing, sizedProp, secure, propBoolean}
 import Arbitrary._
 import Shrink._
 import java.util.Date
+import scala.util.{Try, Success, Failure}
 
 object GenSpecification extends Properties("Gen") {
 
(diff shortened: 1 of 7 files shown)


rickynils:
@non I haven't looked through everything yet, but so far it looks great! Do you think it would be possible to split it up a bit, though? For example, the improvements you've made to `Choose`, could it be a separate commit? On the surface it looks like an isolated change, but I might have missed things.

non:
@rickynils Sure! I can split it out into two parts (although it is related, since the old `Choose` classes would use empty generators when the bounds were impossible).

non:
@rickynils If you want I can close this PR and try to redo it with 2-3 atomic commits (that each pass MiMA/tests).

rickynils:
@non Yeah, that would be awesome, if you're up for it

non:
OK, will do. I'll leave this open now and will close it when I open the new one.

non:
Closed in favor of #258.

PR #261, opened 2016-09-07
Scala 2.12.0-RC1

xuwei-k:


non:
The error appears to be transient.

coltfred:
:+1:  I'd love to get this in so we could start checking other libs against 2.12.0-RC1

rickynils:
I've published Scala 1.12.0-RC1 builds of ScalaCheck 1.11.6, 1.12.5 and 1.13.2

BennyHill:
@rickynils I guess those releases will not include @non 's recent additions that are needed for cats. If so, would it be possible to do a 1.13.3 release? @non I'm assuming the cats issues are indeed covered in your recent PR's here?

non:
@BennyHill I think Cats can work with the previous releases. The changes are nice-to-have but not essential.

BennyHill:
@non But I thought your changes were needed as cats cannot use 1.3.X without them. So this (I think) the required change to get cats on scala 2.12.0-XX

non:
@BennyHill If you ignore the red X (for test coverage reasons -- it has some unused `Cogen` instances) this PR has Cats working with ScalaCheck 1.13.2, ScalaTest 3.0.0, and would support 2.12.0-RC1 as long as its other dependencies all did too.

BennyHill:
@non Awesome! 

And FTR, thanks again to @rickynils  for all this, too!!!

PR #264, opened 2016-09-14
String generators added.

Michaelt293:
* Added the string generators `alphaUpperStr`, `alphaLowerStr` and `alphaNumStr`.
* Changed the order of string generators to match that of the character generators.
* Added tests for `numStr`, `alphaUpperStr`, `alphaLowerStr`, `alphaStr` and `alphaNumStr`.
* Changed `_.forall(c => c.isLetter || c.isDigit)` to `_.forall(_.isLetterOrDigit)` in the definition of `identifier`.

### DIFF
--- a/src/main/scala/org/scalacheck/Gen.scala
+++ b/src/main/scala/org/scalacheck/Gen.scala
@@ -649,27 +649,27 @@ object Gen extends GenArities{
   def identifier: Gen[String] = (for {
     c <- alphaLowerChar
     cs <- listOf(alphaNumChar)
-  } yield (c::cs).mkString).suchThat(_.forall(_.isLetterOrDigit))
+  } yield (c::cs).mkString)
 
   /** Generates a string of digits */
   def numStr: Gen[String] =
-    listOf(numChar).map(_.mkString).suchThat(_.forall(_.isDigit))
+    listOf(numChar).map(_.mkString)
 
   /** Generates a string of upper-case alpha characters */
   def alphaUpperStr: Gen[String] =
-    listOf(alphaUpperChar).map(_.mkString).suchThat(_.forall(_.isUpper))
+    listOf(alphaUpperChar).map(_.mkString)
 
   /** Generates a string of lower-case alpha characters */
   def alphaLowerStr: Gen[String] =
-      listOf(alphaLowerChar).map(_.mkString).suchThat(_.forall(_.isLower))
+      listOf(alphaLowerChar).map(_.mkString)
 
   /** Generates a string of alpha characters */
   def alphaStr: Gen[String] =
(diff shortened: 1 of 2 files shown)


src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -649,15 +649,27 @@ object Gen extends GenArities{
   def identifier: Gen[String] = (for {
     c <- alphaLowerChar
     cs <- listOf(alphaNumChar)
-  } yield (c::cs).mkString).suchThat(_.forall(c => c.isLetter || c.isDigit))
+  } yield (c::cs).mkString).suchThat(_.forall(_.isLetterOrDigit))
+
+  /** Generates a string of digits */
+  def numStr: Gen[String] =
+    listOf(numChar).map(_.mkString).suchThat(_.forall(_.isDigit))
```
xuwei-k: I think we should avoid `suchThat` https://github.com/rickynils/scalacheck/issues/256#issuecomment-244584016

Michaelt293:
@xuwei-k I thought that the use of `suchThat` was redundant for these string generators but I included it since the existing string generators used `suchThat`. Would you prefer the following? If so, I can push these changes.
```scala
  def identifier: Gen[String] = (for {
    c <- alphaLowerChar
    cs <- listOf(alphaNumChar)
  } yield (c::cs).mkString)

  /** Generates a string of digits */
  def numStr: Gen[String] =
    listOf(numChar).map(_.mkString)

  /** Generates a string of upper-case alpha characters */
  def alphaUpperStr: Gen[String] =
    listOf(alphaUpperChar).map(_.mkString)

  /** Generates a string of lower-case alpha characters */
  def alphaLowerStr: Gen[String] =
      listOf(alphaLowerChar).map(_.mkString)

  /** Generates a string of alpha characters */
  def alphaStr: Gen[String] =
    listOf(alphaChar).map(_.mkString)

  /** Generates a string of alphanumerical characters */
  def alphaNumStr: Gen[String] =
    listOf(alphaNumChar).map(_.mkString)
```

non:
@Michaelt293 Yes, I think your proposed change is better. It's possible that a `.suchThat` clause that never fails isn't a big deal, it makes me nervous and can lead to problems later.

Michaelt293:
@non Great. Let me know if I should squash the commits.

rickynils:
The reason for the `suchThat` calls is that this installs a filter that is later used when generated values are shrunk. This helps avoiding some of the confusing cases where you get back a test case that the generator didn't generate from the start. But that whole mechanism is ugly anyway, so I rather have people confused over test case shrinking, than having those `suchThat` calls there.

non:
This looks good to me. 👍

PR #267, opened 2016-09-25
filter properties based on a criteria

ssanj:
Allows properties to be filtered based on a supplied criteria. This
allows running of a subset of property tests for a given specification.
The test criteria is specified by the '-f' option.

Eg:

Given the following properties for a String specification:
1. StringProps.String map
2. StringProps.String reverse

when supplied with a "map" criteria, Scalacheck will only run the map property:

1. StringProps.String map

On the commandline:

scala -cp scalacheck.jar:. StringProps -f "map"

Through SBT:

test-only *StringProps -- -f map

The primary motivation for creating this functionality was so that
I could selectively run a subset of properties that were expensive
to run every time. It also helps focus on the current property being
tested without having to run all properties of a specification each
time.

I'm not sure I have covered all the bases here - I have tested this
through SBT and the commandline, but I wonder if there are any other
scenarios I should test this through.

I'm happy to get some feedback on this and improve it as necessary.

### DIFF
--- a/doc/UserGuide.md
+++ b/doc/UserGuide.md
@@ -723,6 +723,7 @@ Available options:
   -minSuccessfulTests, -s: Number of tests that must succeed in order to pass a property
   -maxDiscardRatio, -r: The maximum ratio between discarded and succeeded tests allowed before ScalaCheck stops testing a property. At least minSuccessfulTests will always be tested, though.
   -maxSize, -x: Maximum data generation size
+  -propFilter, -f: Regular expression to filter properties on
 ```
 
 These command line arguments can also be used in SBT to tweak ScalaCheck's
--- a/jvm/src/test/scala/org/scalacheck/PropertyFilterSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/PropertyFilterSpecification.scala
@@ -1,0 +1,88 @@
+/*-------------------------------------------------------------------------*\
+**  ScalaCheck                                                             **
+**  Copyright (c) 2007-2016 Rickard Nilsson. All rights reserved.          **
+**  http://www.scalacheck.org                                              **
+**                                                                         **
+**  This software is released under the terms of the Revised BSD License.  **
+**  There is NO WARRANTY. See the file LICENSE for the full text.          **
+\*------------------------------------------------------------------------ */
+
+package org.scalacheck
+
+import Prop.BooleanOperators
+
+object PropertyFilterSampleSpecification extends Properties("PropertyFilterSample") {
+
+  property("positive numbers") = Prop.forAll(Gen.posNum[Int]) { n => n > 0 }
+
+  property("negative numbers") = Prop.forAll(Gen.negNum[Int]) { n => n < 0 }
+
+  property("lowercase alpha characters") = Prop.forAll(Gen.alphaLowerChar) { c =>
+    c.toInt >= 97 && c.toInt <= 122
+  }
+}
+
+object PropertyFilterSpecification extends Properties("PropertyFilter") {
+
+  val nl = System.lineSeparator
+
+  private def diff(filter: Option[String], actual: Seq[String],
+    expected: Seq[String]): String = {
+    s"filter: ${filter.getOrElse("not supplied")}" +
+    s"${nl}expected values:$nl" +
+    s"\t${expected.mkString(s"$nl\t")}" +
+    s"${nl}actual values:$nl" +
+    s"\t${actual.mkString(s"$nl\t")}"
+  }
+
+  private def prop(filter: Option[String], actualNames: Seq[String],
+    expectedNames: Seq[String]): Prop = {
+    def lengthProp = actualNames.length == expectedNames.length
+
+    def props = actualNames.forall(expectedNames.contains)
+
+    (lengthProp && props) :| diff(filter, actualNames, expectedNames)
+  }
+
+  property("filter properties by predicate") =
+    Prop.forAllNoShrink(
+      Gen.option(
+        Gen.oneOf(
+          "PropertyFilterSample.*numbers",
+          "PropertyFilterSample.*alpha"))) { pf =>
+
+            val testParams =
+              Test.Parameters.default.
+                withPropFilter(pf)
+
+            val props =
+              Test.checkProperties(
+                testParams,
+                PropertyFilterSampleSpecification
+              )
+
+            val propNames = props.map(_._1)
+
+            if (pf.exists(_.contains("*numbers"))) {
+              val expected =
+                Seq(
+                  "PropertyFilterSample.positive numbers",
+                  "PropertyFilterSample.negative numbers"
+                )
+
+                prop(pf, propNames, expected)
+            } else if (pf.exists(_.contains("*alpha"))) {
+              val expected = Seq("PropertyFilterSample.lowercase alpha characters")
+
+              prop(pf, propNames, expected)
+            } else { //no filter
+              val expected = Seq(
+                  "PropertyFilterSample.positive numbers",
+                  "PropertyFilterSample.negative numbers",
+                  "PropertyFilterSample.lowercase alpha characters"
+                )
+
+              prop(pf, propNames, expected)
+            }
+  }
+}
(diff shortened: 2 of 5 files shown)


non:
Hi @ssanj,

If you're going to do this filtering by matching on test/property names, I think the `-f` argument should accept a regular expression. That means instead of `"map"` you'd use something like `".* map"` instead. Right now, the fact that you can only match against the property name (and not the test name as well) seems like an unnecessary limitation.

What do you think?

ssanj:
Hi @non,

Thanks for your feedback.

1. Matching on regular expression - Good call. I'm not sure why I didn't do this in the first place. It's probably because I based this functionality on similar functionality in ScalaTest (-z option) which uses a [substring ](http://www.scalatest.org/user_guide/using_the_runner) as opposed to a regular expression.
2. Matching against the test name as well as property name - I'm happy to change this to include the test name. The reason I didn't was because, I primarily use this feature from SBT and as such, SBT's test-only task filters the test names for you:

`test-only <test name filter> -- -f <property name filter>` 

We could pass the full match for the test name as well to -f, but that would only apply to what SBT hands over to ScalaCheck after filtering on test name. I thought this might be confusing to use as sometimes based on your SBT test name filter, your property name would not match any tests (because SBT has already filtered it out)

`test-only *StringProp -- --f "List.* map"`

When running Scalacheck from the command line, we only specify a single test (currently - to mu knowledge) so I assumed there was no need to filter by test name. I'm not sure if there's a way to specify a wildcard for test name here?

`scala -cp scalacheck.jar:. StringProps`

So if we match on test name as well and run the filter from the command line, would we have to enumerate all  the tests for the current project and then apply the test and property name filter to those found? I'm not sure what we would specify as the test name in this case?

`scala -cp scalacheck.jar:.  <what would we specify here?> -f "String.* map"`

One option would be to make the test name argument optional when specified with the -f option. Or we could use a wildcard:

`scala -cp scalacheck.jar:.  * -f "String.* map"`

What are your thoughts on this?

non:
@ssanj Ah, I hadn't thought about the fact that `test-only` is already allowing you to select which test(s) you want. Given that, I think using a regex to match the property name itself is fine. The nice thing about that is that it means people can use a suffix (or prefix) to "group" their properties and then run those groups using a regex.

ssanj:
@non cool. Sounds good.

Also I noticed that Travis had failed with some binary compatibility issues:

> [info] scalacheck: found 5 potential binary incompatibilities while checking against org.scalacheck:scalacheck_2.11:1.13.1  (filtered 2)

> [error]  * method copy(Int,Int,Int,Int,org.scalacheck.Test#TestCallback,Float,scala.Option)org.scalacheck.Test#Parameters#cp in class org.scalacheck.Test#Parameters#cp does not have a correspondent in current version

> [error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.scalacheck.Test#Parameters#cp.copy")

> [error]  * method this(org.scalacheck.Test#Parameters,Int,Int,Int,Int,org.scalacheck.Test#TestCallback,Float,scala.Option)Unit in class org.scalacheck.Test#Parameters#cp does not have a correspondent in current version

> [error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.scalacheck.Test#Parameters#cp.this")

> [error]  * the type hierarchy of object org.scalacheck.Test#Parameters#cp is different in current version. Missing types {scala.runtime.AbstractFunction7}

> [error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("org.scalacheck.Test$Parameters$cp$")

> [error]  * method apply(Int,Int,Int,Int,org.scalacheck.Test#TestCallback,Float,scala.Option)org.scalacheck.Test#Parameters#cp in object org.scalacheck.Test#Parameters#cp does not have a correspondent in current version

> [error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("org.scalacheck.Test#Parameters#cp.apply")

> [error]  * abstract method runFilter()scala.Option in class org.scalacheck.Test#Parameters is present only in current version

>[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.scalacheck.Test#Parameters.runFilter")

> java.lang.RuntimeException: scalacheck: Binary compatibility check failed!`

As I've added a new parameter to Test#Parameters and updated cp in the process, should I add this PR against another version branch?

non:
@ssanj As far as MiMA goes let's wait and see what @rickynils thinks.

rickynils:
This looks useful! I think we should simply do 1.13.3 release, and then bump the master branch for 1.14 development. Then we can merge this and #263. I usually maintain a 1.13.x branch with back-ported stuff if needed, rather than creating a new branch for the next release.

.gitignore
```diff
@@ -18,3 +18,7 @@ target
 
 /examples/simple-sbt/project/
 /examples/simple-sbt/target/
+
+# ide
+*.ctags_srcs/
+.tags
```
rickynils: @ssanj Maybe those gitignore changes could be moved to another commit/PR?

ssanj: @rickynils  sure.

ssanj:
@non I've added in support for filtering on the Properties name as well as the property name as you requested before. I found some examples in ScalaCheck that use different test classes/specifications to define properties across the same Properties name. In those cases, SBT only filters on the class/specification name, which is different to the Properties name.

I've added some thoughts on why in the comments to 3d5d030f9866218cdbec944bc737d12acb6f909e.

ssanj:
Seems like [Travis ](https://travis-ci.org/rickynils/scalacheck/jobs/163422722) is having issues downloading dependencies:

> :: problems summary ::
> 
> :::: WARNINGS
> 
>         [FAILED     ] org.scala-sbt#main;0.13.9!main.jar: The HTTP response code for https://repo.typesafe.com/typesafe/ivy-releases/> org.scala-sbt/main/0.13.9/jars/main.jar did not indicate a success. See log for more detail. (345ms)
> 
>         [FAILED     ] org.scala-sbt#main;0.13.9!main.jar: The HTTP response code for https://repo.typesafe.com/typesafe/ivy-releases/> org.scala-sbt/main/0.13.9/jars/main.jar did not indicate a success. See log for more detail. (345ms)
> 
>     ==== typesafe-ivy-releases: tried
> 
>       https://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/main/0.13.9/jars/main.jar
> 
>         [FAILED     ] org.scala-sbt#compiler-interface;0.13.9!compiler-interface-bin.jar: The HTTP response code for https://> repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/compiler-interface/0.13.9/jars/compiler-interface-bin.jar did not > indicate a success. See log for more detail. (362ms)
> 
>         [FAILED     ] org.scala-sbt#compiler-interface;0.13.9!compiler-interface-bin.jar: The HTTP response code for https://> repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/compiler-interface/0.13.9/jars/compiler-interface-bin.jar did not > indicate a success. See log for more detail. (362ms)
> 
>     ==== typesafe-ivy-releases: tried
> 
>       https://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/compiler-interface/0.13.9/jars/compiler-interface-bin.jar
> 
>         ::::::::::::::::::::::::::::::::::::::::::::::
> 
>         ::              FAILED DOWNLOADS            ::
> 
>         :: ^ see resolution messages for details  ^ ::
> 
>         ::::::::::::::::::::::::::::::::::::::::::::::
> 
>         :: org.scala-sbt#main;0.13.9!main.jar
> 
>         :: org.scala-sbt#compiler-interface;0.13.9!compiler-interface-bin.jar
> 
>         ::::::::::::::::::::::::::::::::::::::::::::::
> 
> :: USE VERBOSE OR DEBUG MESSAGE LEVEL FOR MORE DETAILS
> 
> download failed: org.scala-sbt#main;0.13.9!main.jar
> 
> download failed: org.scala-sbt#compiler-interface;0.13.9!compiler-interface-bin.jar
> 
> Error during sbt execution: Error retrieving required libraries
> 
>   (see /home/travis/.sbt/boot/update.log for complete log)
> 
> Error: Could not retrieve sbt 0.13.9>

non:
@ssanj I think you can re-run Travis by closing and reopening (or @rickynils can re-run it directly).

ssanj:
Closing (and will then reopen) to rerun Travis.

ssanj:
@non The Binary compatibility checks still fail as I have changed the API. The other tests have passed.

non:
Looks good, thanks!

PR #268, opened 2016-09-26
Allow type classes to be serializable

alexarchambault:
This PR allows `Arbitrary`, `Gen`, `Cogen`, `Shrink` instances, to be `Serializable`, so that these can be used in contexts relying on Java serialization (from closures with Spark in particular).

The unit tests in this PR are far from exhaustive, but illustrate that simple type class instances are indeed serializable.

Note that scalacheck relying on `CanBuildFrom` at some places tends to break serializability by default, as default `CanBuildFrom` instances from the standard library are usually not serializable. To circumvent that, this PR adds a few serializable `CanBuildFrom` instances (in `SerializableCanBuildFroms`) that get implicitly found if `SerializableCanBuildFroms._` is imported.


non:
Seems reasonable to me. 👍

PR #270, opened 2016-09-27
Fix invalid bounds on `Gen.posNum` (and `Gen.negNum`).

### LINKED ISSUE #269: posNum throws an IllegalBoundsError in 1.13.3-SNAPSHOT
If I run the following on scalacheck 1.13.2:

```
import org.scalacheck._
Prop.forAll(Gen.posNum[Int]) { n => n > 0 }.check
```

I get:
> + OK, passed 100 tests.

But with the 1.13.3-SNAPSHOT, I get:

> ! Exception raised on property evaluation.
> Exception: org.scalacheck.Gen$Choose$IllegalBoundsError: invalid bounds:
  low=1, high=0
> org.scalacheck.Gen$Choose$$anon$5.choose(Gen.scala:331)
> org.scalacheck.Gen$Choose$$anon$5.choose(Gen.scala:329)
> org.scalacheck.Gen$Choose$$anon$7.choose(Gen.scala:362)
> org.scalacheck.Gen$$anonfun$posNum$1.apply(Gen.scala:682)
> org.scalacheck.Gen$$anonfun$posNum$1.apply(Gen.scala:682)

Has some functionality changed around this area? Or am I incorrectly using posNum?

non:
This fixes #269. It also includes a (possibly unnecessary) change to
make `Gen.buildableOf` a bit safer, as well as regression tests.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -358,4 +358,10 @@ object GenSpecification extends Properties("Gen") {
     }
   }
   ////
+
+  property("positive generators are positive #269") =
+    Prop.forAll(Gen.posNum[Int]) { n => n > 0 }
+
+  property("negative generators are negative") =
+    Prop.forAll(Gen.negNum[Int]) { n => n < 0 }
 }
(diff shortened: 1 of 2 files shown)


non:
@ssanj Want to review this? It fixes the issue you noticed.

More generally, the change is working as intended: there were invalid bounds being used, we caught them, and were able to go back to the source (`posNum`) and fix the issue.

One thing this says to me is that maybe I need to add more unit tests for all the different generators defined in `Gen` (and possibly introduce code coverage metrics to see how we're doing so far).

non:
These Travis failures seem like resolution/downloading failures. I don't think they apply to the actual code change in question.

ssanj:
@non Sure. I'll have a look this evening. I have a project that has failing tests because of this bug. Should be easy to see if it all works with the fix.

PR #271, opened 2016-09-28
0xFFFF Is also not unicode.

cogpp:
This issue came up generating strings for use in Xml; code-point 0xFFFF is for internal use only see http://bit.ly/2cVBrzK. Since the code already excludes the other internal character 0xFFFE it seems to make sense to include this one too.


non:
Looks like Travis failed due to resolution issues.

This change looks good to me. 👍

cogpp:
Just closing and re-opening to re-trigger travis.

PR #272, opened 2016-10-07
CogenSpecification

adamgfraser:
This proposal creates a new test suite for Cogen. The tests are based on the concept that Cogen instances should satisfy two properties:

**1. Consistency** - A given Cogen should always return the same output for the same seed and input. In other words, Cogen is referentially transparent.
**2. Uniqueness** - Given a seed, a Cogen should always return different outputs for inputs that are not equal. One can imagine a Cogen that does not do this (e.g. one that is only based on the length of a string to a model a function that only uses that information), but especially for Cogen instances that will generally be available implicitly this seems like a highly desirable property.

With a few minor provisions, all of the existing Cogen instances satisfy these laws. In the case of Function0, Array, Throwable, and Exception we need to use a different definition of equality than the standard `==` but these are fine. For BigDecimal there is a corner case where `BigDecimal(0, Int.MaxValue) == BigDecimal(0, Int.MinValue)` evaluates to true but the Cogen instances returns different values. Not quite sure how to handle that.

I also made a couple of minor changes to the code for Cogen itself. First, I removed the call to `.next` on constructing a new Cogen. I don't think this is necessary and by removing it we can establish several other nice laws for Cogen. Specifically, `contramap(a => a)` is now an identity function consistent with the laws for contravariant functors. Also, we can show that Cogen forms a Monoid and a Divisible (the contravariant equivalent of an Applicative). Hopefully this is not too much category theory but this does provide some additional ways to compose Cogen if someone wants to go down the road of creating more sophisticated arbitrary functions and provides some laws to reason about them.

This is my first pull request so open to any and all feedback. Hopefully this is helpful.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/CogenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/CogenSpecification.scala
@@ -1,0 +1,150 @@
+package org.scalacheck
+
+import org.scalacheck.Arbitrary.arbitrary
+import org.scalacheck.Gen.listOfN
+import org.scalacheck.GenSpecification.arbSeed
+import org.scalacheck.Prop.forAll
+import org.scalacheck.rng.Seed
+
+import scala.util.Try
+
+object CogenSpecification extends Properties("Cogen") {
+
+  // We need a customized definition of equality for some of our tests.
+  trait Equal[A] {
+    def equal(a1: A, a2: A): Boolean
+  }
+
+  object Equal extends EqualLowPriority {
+
+    def apply[A: Equal]: Equal[A] = implicitly
+
+    // Value equality for arrays as opposed to reference equality, consistent with other sequences.
+    implicit def arrayEqual[A: Equal]: Equal[Array[A]] = new Equal[Array[A]] {
+      override def equal(a1: Array[A], a2: Array[A]) =
+        a1.corresponds(a2)(Equal[A].equal)
+    }
+
+    // Two thunks are equal if evaluating them results in the same value.
+    implicit def function0Equal[A: Equal]: Equal[() => A] = new Equal[() => A] {
+      override def equal(a1: () => A, a2: () => A) =
+        Equal[A].equal(a1(), a2())
+    }
+
+    // Two exceptions are equal if they have the same string representation.
+    implicit val exceptionEqual = new Equal[Exception] {
+      override def equal(a1: Exception, a2: Exception): Boolean =
+        a1.toString == a2.toString
+    }
+
+    // Two throwables are equal if they have the same string representation.
+    implicit val throwableEqual = new Equal[Throwable] {
+      override def equal(a1: Throwable, a2: Throwable): Boolean =
+        a1.toString == a2.toString
+    }
+
+    // Two cogens are equal if for all combinations of seeds and values they product the same result.
+    implicit def cogenEqual[A: Arbitrary]: Equal[Cogen[A]] = new Equal[Cogen[A]] {
+      def equal(a1: Cogen[A], a2: Cogen[A]): Boolean =
+        listOfN(100, arbitrary[(Seed, A)]).sample.get.forall {
+          case (x, y) => a1.perturb(x, y) == a2.perturb(x, y)
+        }
+    }
+  }
+
+  // Avoid reimplementing equality for other standard classes.
+  trait EqualLowPriority {
+    implicit def universal[A] = new Equal[A] {
+      override def equal(a1: A, a2: A): Boolean = a1 == a2
+    }
+  }
+
+  // A version of distinct that accepts a custom notion of equality.
+  def distinct[A: Equal](as: List[A]): List[A] =
+    as.foldLeft(List.empty[A])((b, a) => if (b.exists(Equal[A].equal(a, _))) b else a :: b)
+
+  implicit def arbFunction0[A: Arbitrary]: Arbitrary[() => A] =
+    Arbitrary(arbitrary[A].map(() => _))
+
+  implicit def arbCogen[A: Arbitrary : Cogen]: Arbitrary[Cogen[A]] =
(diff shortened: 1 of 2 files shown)


non:
This looks great, thanks! 👍 

It's incredibly unlikely to fail in practice, but do you want to put in a comment for the uniqueness law which says that this is only required *approximately*? That is, for types whose space of possible values is larger than the space of the seed (2^256), it's possible that there will be legitimate collisions (but very unlikely for given the number of samples we're testing).

Also, do you think it's worth "fixing" the `Cogen[BigDecimal]` instance?

adamgfraser:
Yes, added a comment to reflect that.

Regarding the BigDecimal instance, yes I think it is worth fixing and as I got into it there were a couple of complications. The previous implementation was based on perturbing a pair of the unscaled value and the scale. This created two issues:

- When the unscaled value is zero the scaling factor is irrelevant so the Cogen instance was creating different results for equal inputs.
- BigDecimal does not enforce consistency in the way that unscaled values and scales are used in creating instances. For example, `BigDecimal(30)` and `BigDecimal(3, -1)` are equal but have different combinations of unscaled values and scales.

I added code to deal with the zero unscaled value case and to normalize BigDecimal values to deal with the second issue.

non:
Looks great!

One last thing: I think it might be better to write normalize as something like:

```scala
def normalize(unscaled: BigInteger, scale: Int): (BigInteger, Int) = ???
```

This way you can just use `divideAndRemainder` directly, avoid constructing a bunch of intermediate stuff, etc. It's a bit of a bikeshed, but I think having this code be more efficient (especially for very large values) will be appreciated by users.

What do you think?

adamgfraser:
Yes, makes a ton of sense. Just refactored it along those lines.

non:
Looks great. 👍 

Seems like once #276 is merged the tests will all pass and we can merge this too. Thanks again!

adamgfraser:
Hmmm, interesting. I didn't get that error message when I ran the tests locally. Hopefully making Cogen invariant will resolve it. I'm trying to get more involved with community projects so let me know if there is anything else you're looking for help with.

non:
Sure thing. FWIW the PRs look great.

I think @rickynils is the only person who can merge a PR, so we'll both have to wait until he can take a look.

jvm/src/test/scala/org/scalacheck/CogenSpecification.scala
```diff
@@ -0,0 +1,223 @@
...
+    def apply[A: Equal]: Equal[A] = implicitly
+
+    // Value equality for arrays as opposed to reference equality, consistent with other sequences.
+    implicit def arrayEqual[A: Equal]: Equal[Array[A]] = new Equal[Array[A]] {
+      override def equal(a1: Array[A], a2: Array[A]): Boolean =
+        a1 sameElements a2
+    }
+
+    // Two thunks are equal if evaluating them results in the same value.
+    implicit def function1Equal[A: Equal]: Equal[() => A] = new Equal[() => A] {
```
xuwei-k: function0 ?

adamgfraser:
Thanks! Just fixed.

rickynils:
@adamgfraser Thanks for this! Could you clean up the commits a bit? For example rename the one named `First commit` and remove that pair of contravariance commit+revert.

adamgfraser:
Cleaned this one up as well.

rickynils:
@adamgfraser Could I ask for yet another rebase against master now that I've merged #276? Just so we'll see that green Travis dot before merging this :)

adamgfraser:
No problem. Just did it.

non:
@adamgfraser Looks like the binary compatibility check is failing here. I think adding `divide` and `divided` will need to wait until 1.14.0 :/

Sorry about this! Scala's traits are really brittle -- I wish I had used `abstract class` here, since then adding methods would still be compatible.

adamgfraser:
No problem. Shall I revise my pull request to remove those? I could also just move them to the companion object. It looks like the methods I added there aren't creating binary compatibility issues.

non:
It's your call. I think you should either remove them or move them, and then rebase and push again.

Since these methods weren't previous used, my preference would be to just remove them then add them once we are ready for 1.14.0, since ultimately that gets the API to where we'd want it in the longterm.

adamgfraser:
Okay, I will take them out.

One more thing. As I was going through this I added tests for tuple2Cogen and function1Cogen and noticed that function1Cogen is violating our laws. For my tests I'm defining two functions as equal if they generate identical outputs for 100 arbitrary inputs. The issue is that for our Cogen instance we're generating a perturbed seed based on the output of only one arbitrarily chosen input, so if f(x) and g(x) generate the same output for x = 2 and 2 is chosen, we will generate the same Cogen output, even though f and g may be different functions that produce different outputs for many other inputs. We could potentially address by generating a set of random values for the input, evaluating the function with all of them, and then perturbing based on all the outputs we get. Worthwhile or too much?

non:
@adamgfraser So this is where my argument about the law being *approximate* comes in. I would say that if you want to get these instances passing, a better thing to do is to rewrite the cogen law so that it uses different seeds (instead of just one seed). This makes it much less likely that something like this will happen.

I don't think we should change the internal working of `Cogen[A => B]` just to get this law passing (especially since this failure isn't catching an actual problem, just a limitation in our testing).

adamgfraser:
@non Good point. I tried a couple of variations of the law along the lines you suggested such as:

`def uniqueness_[A: Equal : Arbitrary : Cogen]: Prop =  forAll { (x: A, y: A, seed: Seed) =>     Equal[A].equal(x, y) == (Cogen[A].perturb(seed, x) == Cogen[A].perturb(seed, y))   }`

I was still getting issues with those, but as you say the uniqueness is only approximate. I went ahead and removed the methods that were causing the binary compatibility issue, so pending the green light from Travis this should be good to go. Thanks for your help and feedback along the way.

non:
👍 

Thanks for sticking with this! This looks great.

PR #277, opened 2016-10-17
Arbitrary Function0 Instance

adamgfraser:
Added an Arbitrary instance for Function0. This is super simple but I found myself writing it for something and thought it would be good to include so others have it off the shelf.

### DIFF
--- a/src/main/scala/org/scalacheck/Arbitrary.scala
+++ b/src/main/scala/org/scalacheck/Arbitrary.scala
@@ -58,7 +58,12 @@ sealed abstract class Arbitrary[T] extends Serializable {
  *  generators.
  *  </p>
  */
-object Arbitrary extends ArbitraryLowPriority with ArbitraryArities
+object Arbitrary extends ArbitraryLowPriority with ArbitraryArities {
+
+  /** Arbitrary instance of the Function0 type. */
+  implicit def arbFunction0[T](implicit a: Arbitrary[T]): Arbitrary[() => T] =
+  Arbitrary(arbitrary[T] map (() => _))
+}
 
 /** separate trait to have same priority as ArbitraryArities */
 private[scalacheck] sealed trait ArbitraryLowPriority {


non:
👍

non:
Unfortunately adding this will break binary compatibility (due to the fragility of traits).

There are some other `Arbitrary` instances we'd like to add (for example, `Symbol`). Once we're able to merge this let's make sure we hit all those too.

(Also, @rickynils, once we break compatibility to merge this, is it worth changing the trait in question to be an abstract class to avoid this in the future?)

rickynils:
@adamgfraser Some rebasing needed here too :) In general, I recommend looking over the commit history when you're think you're done with a PR. Make sure that each commit has a reasonably describing name, and if there maybe are stuff that could be squashed into fewer commits to make it more logical and easier to follow. Once the PR is merged, that commit history will end up directly on the ScalaCheck master branch, and it can be tricky to follow the changes if something needs to be debugged or reverted.

@non Yeah, changing to abstract class sounds reasonable.

adamgfraser:
Thanks for the feedback! Still learning this stuff so really appreciate it. Just rebased this one and will go back and clean up the others. Would you please confirm that this is what you are looking for?

rickynils:
@adamgfraser Yeah, exactly what I was looking for! Thanks!

xuwei-k:
I think we can keep binary compatibility if add to the `object Arbitrary`

```diff
- object Arbitrary extends ArbitraryLowPriority with ArbitraryArities
+ object Arbitrary extends ArbitraryLowPriority with ArbitraryArities {
+  
+  /** Arbitrary instance of the Function0 type. */
+  implicit def arbFunction0[T](implicit a: Arbitrary[T]): Arbitrary[() => T] =
+    Arbitrary(arbitrary[T] map (() => _))
+ }
```

adamgfraser:
@xuwei-k Thanks for the suggestion! I made the change and it passed all tests locally. Just pushed it so we will see if it passes binary compatibility with Travis. I guess the other question would be how we feel from a design perspective putting it there. It is a little weird that arbFunction0 would now be the highest priority arbitrary instance in the project, but that is a little more of an aesthetic thing as it is a very specific type so don't think there is a risk of collisions.

@rickynils Do you like this solution?

adamgfraser:
Alright, well looks like it works. If people like this I can clean up the commit a bit and then it should be good to go.

non:
Give this a rebase and then 👍

adamgfraser:
Just rebased.

non:
👍

PR #278, opened 2016-10-17
Fix non-terminating shrink streams.

### LINKED ISSUE #244: Simple check leads to infinite loop
Hi, 

I just started playing around with ScalaCheck and encountered a problem. The following does not terminate on my machine with ScalaCheck 1.13.[0|1]:

```scala
import org.scalacheck.Prop.forAll

object HelloSpec extends Properties("Hello") {

  def replaceLettersWithX(str: String) = {

    def replaceLetter(ch: Char): Char = {
      if (ch.isLetter) {
        'X'
      } else {
        ch
      }
    }

    str map replaceLetter
  }

  property("is actually the identity function") = forAll { (str: String) =>
    val withX: String = replaceLettersWithX(str)
    withX == str
  }

}
```

With version 1.12.5, a counter-example is produced quickly. With the newer versions, it gets stuck in an infinite loop. I tried to debug it and it seems like once ScalaCheck has found a counter-example, it is trying to shrink it. The shrinking however only keeps replacing one 1-character string with another.

non:
This was reported as #244. The root cause was that shrinking certain
values would produce an infinite stream. There were several ways this
could happen -- in the reported case, the fact that Char was unsigned
meant that the strategy of including inverses was able to produce
loops.

The fix was to detect when Integral[T] cannot support negative values,
and disabling negation.

While working on this, I noticed that shrinking floating point
sentinels (i.e. infinities and NaN) would cause a similar issue. I
added code to detect those cases as well.

Fixes #244.

### DIFF
--- a/src/main/scala/org/scalacheck/Shrink.scala
+++ b/src/main/scala/org/scalacheck/Shrink.scala
@@ -226,15 +226,22 @@ object Shrink extends ShrinkLowPriority {
 }
 
 final class ShrinkIntegral[T](implicit ev: Integral[T]) extends Shrink[T] {
-  import ev.{ equiv, fromInt, zero, minus, times, quot }
-  val minusOne = fromInt(-1)
+  import ev.{ fromInt, lt, gteq, quot, negate, equiv, zero, one }
+
   val two = fromInt(2)
 
-  // assumes x is non-zero
+  // see if T supports negative values or not. this makes some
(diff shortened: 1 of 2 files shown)


non:
(This issue was first introduced with the introduction of the generic shrinker in 1.13.0.)

rickynils:
Thanks!

Issue #280, opened 2016-10-17
Add new Gen/Arbitrary instances


non:
Currently adding these breaks binary compatibility. This issue is intended to catalogue all the additional instances we want to add (`Gen`, `Cogen`, and `Arbitrary`).

We should also change the trait containing these instances to an `abstract class`. This way future additions won't create binary incompatibilities.

Here are the instances I can think of:
- [ ] `Function0` instances (#277)
- [ ] `Symbol` instances (#243)
- [ ] `java.math.BigInteger`
- [ ] `java.math.BigDecimal`
- [ ] `java.time.*` instances for Scala 2.12/Java 8 (#239)
- [ ] `Gen.infiniteStream`

(I'll update this list as necessary.)

adamgfraser:
I was thinking of adding a Gen instance for infinite streams. Would there be interest in that?

non:
@adamgfraser Sure, I think a `Gen.infiniteStream[A]` would be really useful.

adamgfraser:
Would people be interested in a `Gen.bigList` instance? I'm thinking that sometimes it can be valuable to generate a random large collection to check the stack safety of a method. You can already do this today with something like `Gen.listofN(10000, arbitrary[A])` but thought it might be a nice convenience to have a generator that produced a mix of normal lists, really big lists, and empty lists similar to what already exists for numbers so users can test multiple cases with one property. This could be implemented as just an additional Gen instance so it wouldn't impact any existing use cases.

At the same time I understand that there is value in not clogging up the API with too many specific use cases as opposed to more general combinators so open to whatever people think is best.

non:
Personally I feel like it's easy enough to use `listOfN` here. I could also imagine someone else publishing a third-party library with a bunch of extra handy `Gen` instances/builders.

non:
@rickynils I just noticed many of the binary-incompatible PRs got merged. Does this mean *master* is ready for these kinds of changes?

rickynils:
@non Yes, master is bumped to 1.14 so we can bring on those changes. However, we should also think about what to do with this feedback: https://twitter.com/extempore2/status/788984845913337856

non:
OK, yeah. I'll give it some thought.

kevinmeredith:
Out of curiosity, why is `java.math.BigDecimal` in your list, @non, but not `scala.math.BigDecimal`?

non:
[...]

PR #281, opened 2016-10-18
Gen.infiniteStream

adamgfraser:
Adds a Gen instance for generating an infinite stream of values given a generator for the type of the underlying values. This is useful for testing that functions designed to be able to handle infinite data structures (e.g. through lazy evaluation) are working correctly. Also includes a new test for this instance. See #280.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -161,6 +161,10 @@ object GenSpecification extends Properties("Gen") {
     l.length == 0
   }
 
+  property("infiniteStream") = forAll(infiniteStream(arbitrary[Int]), arbitrary[Short]) { (s, n) =>
+    s.drop(n & 0xffff).nonEmpty
+  }
+
   property("someOf") = forAll { l: List[Int] =>
     forAll(someOf(l))(_.toList.forall(l.contains))
   }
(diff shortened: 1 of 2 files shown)


src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -574,6 +574,21 @@ object Gen extends GenArities{
    *  is equal to calling <code>containerOfN[Map,T,U](n,g)</code>. */
   def mapOfN[T,U](n: Int, g: Gen[(T,U)]) = buildableOfN[Map[T,U],(T,U)](n,g)
 
+  private[scalacheck] def unfold[A, S](z: S)(f: S => Option[(A, S)]): Stream[A] =
```
adamgfraser: Thanks! I thought about this too and just made the change.

jvm/src/test/scala/org/scalacheck/GenSpecification.scala
```diff
@@ -161,6 +161,10 @@ object GenSpecification extends Properties("Gen") {
     l.length == 0
   }
 
+  property("infiniteStream") = forAll(infiniteStream(arbitrary[Int]), arbitrary[Int => Boolean]) { (s, p) =>
+    s exists p
+  }
```
adamgfraser: Yes, I was probably getting a little cute here. I think the likelihood of the old test hanging is incredibly remote given the way arbitrary functions are generating and that is the only case where it wouldn't work. But at the same time it would be incredibly likely to pass for even finite streams with a length of 100, so it isn't like it was a particularly great test. I like the test you suggested and changed it along those lines.

adamgfraser:
On the bright side, from the previous Travis checks it looks like there are no binary compatibility issues so we don't have to wait on this once we get the green light if everyone is good with the changes.

Issue #232, opened 2016-04-08
Stack overflow caused by generator


dabd:
The following generator causes a stack overflow error:

`def bigDecimalGen(lower: BigDecimal = BigDecimal(0), upper: BigDecimal = BigDecimal("999999999999"), withDecimals: Boolean = true): Gen[BigDecimal] =
  for {
    i <- arbBigInt.arbitrary.retryUntil(x => x >= lower.toBigInt && x <= upper.toBigInt)
    d <- if (withDecimals) Gen.chooseNum(0, 99) else Gen.const(0)
  } yield {
    BigDecimal(s"$i.$d").setScale(2)
  }

val fooGen: Gen[(BigDecimal, BigDecimal, BigDecimal)] = for {
  t <- bigDecimalGen(lower = BigDecimal(80000), upper = BigDecimal(150000), withDecimals = false)
  a <- bigDecimalGen(lower = BigDecimal(8000), upper = BigDecimal(20000), withDecimals = false)
  p <- Gen.frequency(
    (50, bigDecimalGen(withDecimals = false).retryUntil(_ <= t)),
    (50, bigDecimalGen(withDecimals = false).retryUntil(_ > t)))
} yield (p, a, t)

fooGen.sample`

will cause 
`java.lang.StackOverflowError
	at org.scalacheck.Gen$$anon$6.doApply(scratch_34:182)
	at org.scalacheck.Gen$$anonfun$map$1.apply(scratch_34:45)
	at org.scalacheck.Gen$$anonfun$map$1.apply(scratch_34:45)
	at org.scalacheck.Gen$$anon$6.doApply(scratch_34:182)
	at org.scalacheck.Gen$$anon$2.doApply(scratch_34:74)
	at org.scalacheck.Gen$$anonfun$sized$1.apply(scratch_34:350)
	at org.scalacheck.Gen$$anonfun$sized$1.apply(scratch_34:350)
	at org.scalacheck.Gen$$anon$6.doApply(scratch_34:182)
	at org.scalacheck.Gen$$anonfun$map$1.apply(scratch_34:45)
	at org.scalacheck.Gen$$anonfun$map$1.apply(scratch_34:45)
	at org.scalacheck.Gen$$anon$6.doApply(scratch_34:182)
	at org.scalacheck.Gen$$anonfun$flatMap$1.apply(scratch_34:49)
	at org.scalacheck.Gen$$anonfun$flatMap$1.apply(scratch_34:48)
[...]

PR #282, opened 2016-10-19
Improve retryUntil's behavior during failure.

non:
Previously retryUntil would throw a StackOverflowException if the
generator failed too many times. It's not obvious to users that the
problem is their generator and/or predicate.

This commit replaces this implicit depth limit with an explicit cutoff
after 10000 tries (along with a custom Exception). It also provides a
two-argument variant of retryUntil that allows the user to specify how
many tries are allowed.

This change should improve the situation in #232.


rickynils:
Thanks! Much better than risking stack overflows

Issue #246, opened 2016-06-09
arbBigDecimal causing Underflow (ArithmeticException)


muhuk:
Sorry for the sloppy report. I'll flesh it out later once I find the time.

I am consistently getting failures like this:

    ArithmeticException was thrown during property evaluation.
      Message: Underflow
      Occurred when passed generated values (
    
      )

(Note that `generated values` is empty.)

I can run my tests if I replace this expression:

    Arbitrary.arbBigDecimal.arbitrary

With this:

    Gen.oneOf(BigDecimal(1.0), BigDecimal(2.0), BigDecimal(3.0))

non:
@muhuk Can you paste a bit more about this error? I'm not sure how to reproduce (and fix) this issue otherwise.

muhuk:
Here is a complete test:

```scala
import org.scalacheck.Arbitrary
import org.scalacheck.Gen
import org.scalatest.Finders
import org.scalatest.WordSpec
import org.scalatest.prop.GeneratorDrivenPropertyChecks

class ArbBigDecimalUnderflow extends WordSpec with GeneratorDrivenPropertyChecks {
  val genBigDecimal = Arbitrary.arbBigDecimal.arbitrary

  "this one should fail" in {
    forAll(Gen.zip(genBigDecimal, genBigDecimal)) {
      case (a, b) ⇒ assert(a + b == b + a)
    }
  }
}
```


I am using:

- scala 2.11.8
- scalatest 2.2.6
- scalacheck 1.13.2

Any other specifics I can provide?

non:
That's great, thanks!

muhuk:
You're welcome.

non:
@muhuk So my sense is that the `Gen[BigDecimal]` instances ScalaCheck uses right now are basically busted. They generate numbers so big that it's really difficult to do anything with them. For example, it's not clear what anyone is going to do with `9E+1054820097`.

What's worse is that the default generic shrinking code will naïvely divide by 2 to make the numbers small enough to work with. Trying to reach `1e-5` by repeatedly dividing `9E+1054820097` by `2` is not going to be a winning strategy.

In my own code I usually roll my own `Gen[BigDecimal]` instances (often created from a `Double` or similar). I think the right way to fix this is to change how ScalaCheck generates these values to make them at least semi-tractable to work with.

What do you think @rickynils?

non:
(The subtext here is that when I tried to run the given example I didn't get an underflow, but my code just hung.)

[...]

PR #283, opened 2016-10-19
Improve BigInt and BigDecimal generators.

non:
Previously, the BigDecimal generator could create astronomically-large
numbers that were not safe to use in most tests (so large that working
with them would cause seemingly-infinite hangs). This was primarily
due to using incredibly large scales (e.g. 10^2000000000). In practice
most users of BigDecimal are probably not trying to work with numbers
that large.

This commit also tries to even out the overall distributions of both
BigInt and BigDecimal numbers, while ensuring that none of the values
are so large that they will break most interesting properties. It's
relatively easy for users to write custom generators if they want
extremely large numbers.

Related to #246.


adamgfraser:
👍

PR #284, opened 2016-10-21
Add support for Duration and FiniteDuration

ceedubs:
This adds `Gen`, `Arbitrary`, and `Shrink` instances for both `Duration`
and `FiniteDuration`. It also adds a `Choose` instance for
`FiniteDuration`.


ceedubs:
Oh I should probably add `Cogen` as well. I'll do that shortly.

non:
👍 once Travis is green.

PR #286, opened 2016-10-21
Reduce the frequency of None in Option Gen

ceedubs:
The idea here is that if you are generating 100 values, it's probably
not that useful for 50 of them to be the same (None). Having a 50%
chance of None is especially problematic when a data structure has
several nested Option fields, in which case it becomes pretty unlikely
that any non-None values are generated for the innermost Option field.

I've arbitrarily chosen a 10% None frequency here, but feel free to
recommend other values that may be better.


non:
👍 

I think 10% `None` sounds about right. Since by default we generate 100 tests, this makes it very likely to get some nones in there without wasting too many test cases on the same value.

ceedubs:
Closing and reopening the PR to try to get the build to pass again (I think it failed because the series of tubes is under attack).

non:
👍

PR #287, opened 2016-10-21
Use seed.next for Success in Cogen for Try

### LINKED ISSUE #285: Inconsistent use of seed.next in Cogen instances
The `Cogen` instances for `Option` and `Either` both use `seed.next` as an argument to `perturb` when they are delegating through to an underlying `Cogen` instance. See [here](https://github.com/rickynils/scalacheck/blob/3f436ffa218329858d6374cb0f7816a1d21588d9/src/main/scala/org/scalacheck/Cogen.scala#L106-L110).

It seems to me that there is inconsistent behavior here. For example, the `Either` instance uses `seed` when calling `perturb` for `Left`s but `seed.next` when calling `perturb` for `Right`s. Also none of the other instances call `seed.next` when delegating to `perturb` (including the instance for `Try`, which is similar to `Either`).

I'm happy to submit a PR, but I don't think that I understand when `seed.next` should be called (if at all) in these instances. cc @non @rickynils

ceedubs:
Fixes #285 (see discussion there).

### DIFF
--- a/src/main/scala/org/scalacheck/Cogen.scala
+++ b/src/main/scala/org/scalacheck/Cogen.scala
@@ -141,7 +141,7 @@ object Cogen extends CogenArities with CogenLowPriority {
 
   implicit def cogenTry[A: Cogen]: Cogen[Try[A]] =
     Cogen((seed: Seed, x: Try[A]) => x match {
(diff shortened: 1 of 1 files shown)


non:
Great! 👍

adamgfraser:
👍

PR #292, opened 2016-10-31
Upgrade to Scala 2.12.0

fedgehog:


non:
Thanks! Before we do something like this we probably need to figure out a work-around for https://github.com/rickynils/scalacheck/issues/290.

fedgehog:
Ah, didn't see that one. Quite a puzzle it is.

PR #293, opened 2016-10-31
Release for Scala 2.12.0

magicknot:
I also bumped `sbt` version to 0.13.13.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ notifications:
 scala:
   - 2.10.6
   - 2.11.8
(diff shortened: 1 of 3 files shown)


non:
Thanks! Before we do this we probably need to figure out a work-around for https://github.com/rickynils/scalacheck/issues/290.

rickynils:
Fixed in ea4a7746d404bc14bd470b75472c55b8aed59058 and f263ce330b7c79bd113ce5d56932d02b664dd0d8

PR #294, opened 2016-10-31
Fixes #290 (deadlock under 2.12 during property initialization)

### LINKED ISSUE #290: Deadlock with ScalaCheck 1.13.x, scalaz-concurrent, and Scala 2.12.0-RC2
After trying to port [remotely](https://github.com/Verizon/remotely) to Scala 2.12.x, I found that some of the tests involving `scala.concurrent.Future` weren't terminating. The most likely cause is a deadlock. The issue doesn't appear in Scala 2.10.x or 2.11.x, and also not when running ScalaCheck properties through ScalaTest. It only manifests when using the plain ScalaCheck runner and 2.12.0-RC2.

A small reproduction project, including instructions, can be found [here](https://github.com/larsrh/scalacheck-futures-deadlock-2.12).

As of now I'm unsure whether this is a problem with `scalaz.concurrent.Task`, `scala.concurrent.Future`, or ScalaCheck (or a combination thereof). I'm opening this issue so that others who might have the same problem can find it.

mpilquist:
Fixes #290 by making property registration lazy.

Mima passes against 1.13.1 so this is safe to release in 1.13.4.

Review by @rickynils.

### DIFF
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -43,7 +43,7 @@ sealed abstract class Prop extends Serializable {
     for(r1 <- this; r2 <- p) yield f(r1,r2)
 
   /** Convenience method that checks this property with the given parameters
-   *  and reports the result on the console. Should only be used when running 
+   *  and reports the result on the console. Should only be used when running
    *  tests interactively within the Scala REPL.*/
   def check(prms: Test.Parameters): Unit = Test.check(
     if(prms.testCallback.isInstanceOf[ConsoleReporter]) prms
@@ -52,16 +52,16 @@ sealed abstract class Prop extends Serializable {
   )
 
   /** Convenience method that checks this property with the given parameters
-   *  and reports the result on the console. Should only be used when running 
-   *  tests interactively within the Scala REPL. 
+   *  and reports the result on the console. Should only be used when running
+   *  tests interactively within the Scala REPL.
    *
    *  The default test parameters
    *  ([[Test.Parameters.default]]) are used for the check. */
   def check: Unit = check(Test.Parameters.default)
 
   /** Convenience method that checks this property and reports the result
-   *  on the console. Should only be used when running 
-   *  tests interactively within the Scala REPL. 
+   *  on the console. Should only be used when running
+   *  tests interactively within the Scala REPL.
    *
    *  The provided argument should be a function that takes
    *  the default test parameters ([[Test.Parameters.default]])
@@ -299,10 +299,6 @@ object Prop {
 
   /** Create a property from a boolean value */
   def apply(b: Boolean): Prop = if(b) proved else falsified
-
-  /** Create a prop that evaluates the by-name argument on each application. */
-  def suspend(p: => Prop): Prop = apply(prms => p(prms))
-
 
   // Implicits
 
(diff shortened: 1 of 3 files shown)


non:
@mpilquist I'm assuming this change fixes the issue in @larsrh's repo?

mpilquist:
Yep:

```
[info] Compiling 2 Scala sources to /private/tmp/scalacheck-futures-deadlock-2.12/target/scala-2.12.0-RC2/test-classes...
[warn] there were two deprecation warnings (since 7.2); re-run with -deprecation for details
[warn] one warning found
[info] ResponseSuite:
[info] Task
[info] - should be stack safe
[info] + Response.stack safety: OK, proved property.
[info] ScalaCheck
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] ScalaTest
[info] Run completed in 303 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
```

dwijnand:
Does mima pass against 1.13.3?

I assume 1.13.3 was mima tested against 1.13.2, 1.13.2 against 1.13.1, etc. If not it might be good to make sure it is in the future, as scalacheck is extremely cornerstone to the Scala ecosystem dependency graph.

mpilquist:
@dwijnand Good idea. I've run mima a bunch more. Now I've run mima against 1.13.0, 1.13.1, 1.13.2, and 1.13.3. All pass. :)

```
› sbt
[info] Loading global plugins from /Users/mpilquist/.sbt/0.13/plugins
[info] Loading project definition from /Users/mpilquist/Development/oss/third-party/scalacheck/project
[info] Set current project to scalacheck (in build file:/Users/mpilquist/Development/oss/third-party/scalacheck/)
> project jvm
[info] Set current project to scalacheck (in build file:/Users/mpilquist/Development/oss/third-party/scalacheck/)
> mimaReportBinaryIssues
[info] Resolving org.scala-sbt#test-interface;1.0 ...
[info] scalacheck: found 0 potential binary incompatibilities while checking against org.scalacheck:scalacheck_2.11:1.13.3
```

mpilquist:
Oh and 1 final test just to double check MIMA... I compiled the following with ScalaCheck 1.13.3 and then confirmed it ran successfully when launched with a JAR containing this change.

```scala
import org.scalacheck._

object Foo extends Properties("Foo") {
  property("foo") = {
    println("Ran")
    true
  }
}
```

non:
Looks like a transient Travis failure. I'm still 👍 on this.

src/main/scala/org/scalacheck/Prop.scala
```diff
@@ -300,6 +300,9 @@ object Prop {
   /** Create a property from a boolean value */
   def apply(b: Boolean): Prop = if(b) proved else falsified
 
+  /** Create a prop that evaluates the by-name argument on each application. */
+  def suspend(p: => Prop): Prop = apply(prms => p(prms))
```
rickynils: In `Gen`, we call this method `delay`: https://github.com/rickynils/scalacheck/blob/master/src/main/scala/org/scalacheck/Gen.scala#L455. Do you think it would be sensible to use the same name in `Prop`?

rickynils:
@mpilquist I have created a branch named `1.13.x` for handling the 1.13.4 release (since master has bben bumped to 1.14). I don't know if it is possible to switch target branch for a PR in github. If not, I can merge this one manually.

xuwei-k:
>  I don't know if it is possible to switch target branch for a PR in github

https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

rickynils:
Ah, thanks @xuwei-k

PR #298, opened 2016-11-03
Make Gen[Calendar] deterministic.

### LINKED ISSUE #295: Regression in `Gen.calendar`
It would appear that `Gen.calendar` somehow uses the current system time in `1.13.3`. Here's a test cases that exhibits this behaviour:

```scala
import org.scalatest.FunSuite
import org.scalatest.prop.GeneratorDrivenPropertyChecks
import java.util.Date
import org.scalacheck._

class Repr extends FunSuite with GeneratorDrivenPropertyChecks {
   test("reproduce") {
    forAll { (s: String, g: String => Date) =>
      val d1 = g(s)
      Thread.sleep(100)
      val d2 = g(s)

      assert(d1 === d2)
    }
  }
}
```

This consistently fails under `1.13.3`, with dates that are 100ms appart. It systematically passes in `1.12.5`, however.

Here's a [discussion](https://gitter.im/typelevel/general?at=5817a5467b15d16e55cc9cd7) on the subject in gitter.

non:
Previously it was possible for the "current system time" to leak
through into generated Calendar instance (and by extention Date
instances). This was especially painful for generated functions (such
as A => Date) where the function is likely to be evaluated at multiple
different system times.

This PR restructures Gen.calendar to ensure that all instances are
initialized from a generated time (in milliseconds). The code is
cleaned up a bit and a test was added to ensure generation is
deterministic.

Fixes #295.

### DIFF
--- a/src/main/scala/org/scalacheck/Gen.scala
+++ b/src/main/scala/org/scalacheck/Gen.scala
@@ -21,6 +21,8 @@ import scala.collection.immutable.TreeMap
 import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.duration.{Duration, FiniteDuration}
 
+import java.util.{ Calendar, UUID }
+
 sealed abstract class Gen[+T] extends Serializable { self =>
 
   //// Private interface ////
@@ -754,16 +756,15 @@ object Gen extends GenArities{
   //// Misc Generators ////
 
   /** Generates a version 4 (random) UUID. */
-  lazy val uuid: Gen[java.util.UUID] = for {
+  lazy val uuid: Gen[UUID] = for {
     l1 <- Gen.choose(Long.MinValue, Long.MaxValue)
     l2 <- Gen.choose(Long.MinValue, Long.MaxValue)
     y <- Gen.oneOf('8', '9', 'a', 'b')
-  } yield java.util.UUID.fromString(
-    new java.util.UUID(l1,l2).toString.updated(14, '4').updated(19, y)
+  } yield UUID.fromString(
+    new UUID(l1,l2).toString.updated(14, '4').updated(19, y)
   )
 
-  lazy val calendar: Gen[java.util.Calendar] = {
-    import java.util.{ Calendar, Date, TimeZone, Locale }
+  lazy val calendar: Gen[Calendar] = {
     import Calendar._
 
     def adjust(c: Calendar)(f: Calendar => Unit): Calendar = { f(c); c }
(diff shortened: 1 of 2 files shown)


src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -763,115 +763,78 @@ object Gen extends GenArities{
...
+    // default, Calendar.getInstance uses the system time. We always
+    // overwrite it with a determinisitcally-geneated time to be sure
+    // that calendar generation is also deterministic.
+    //
+    // We limit the time (in milliseconds) because extreme values will
+    // cause Calendar.getTime calls to fail. This range is relatively
+    // large but safe:
+    //
+    //   -62135751600000 is 1 CE
+    //    64087186649116 is 4000 CE
```
sritchie-stripe: can't wait for all the Y4K bugs this is going to introduce

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -763,115 +763,78 @@ object Gen extends GenArities{
...
-      l <- Gen.chooseNum(Long.MinValue, Long.MaxValue)
-      now = new Date
-      d = new Date(now.getTime + l)
-      c = Calendar.getInstance()
-    } yield {
-      c.setTimeInMillis(d.getTime)
-      c
-    }
 
-    val calendarBeginningOfDayGen: Gen[Calendar] = for {
```
sritchie-stripe: btw @tixxit all these generators make the `Gen` / `Arbitrary` distinction much clearer.

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -763,115 +763,78 @@ object Gen extends GenArities{
...
+    val endOfDayGen: Gen[Calendar] =
+      calendar.map(c => adjust(c) { c =>
+        c.set(HOUR_OF_DAY, 23)
+        c.set(MINUTE, 59)
+        c.set(SECOND, 59)
+        c.set(MILLISECOND, 59)
+      })
+
+    val firstDayOfYearGen: Gen[Calendar] =
+      for { c <- calendar; y <- yearGen(c) } yield adjust(c)(_.set(y, 0, 1))
```
sritchie-stripe: How about `JANUARY` instead of `0`?

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -763,115 +763,78 @@ object Gen extends GenArities{
...
+
+    val lastDayOfYearGen: Gen[Calendar] =
+      for { c <- calendar; y <- yearGen(c) } yield adjust(c)(_.set(y, 11, 31))
```
sritchie-stripe: `DECEMBER` vs `11`

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -763,115 +763,78 @@ object Gen extends GenArities{
...
-    } yield {
-      c.set(Calendar.HOUR_OF_DAY, 23)
-      c.set(Calendar.MINUTE, 59)
-      c.set(Calendar.SECOND, 59)
-      c.set(Calendar.MILLISECOND, 59)
-      c
+    def moveToNearestLeapDate(c: Calendar, year: Int): Calendar = {
+      @tailrec def loop(y: Int): Calendar = {
+        c.set(YEAR, y)
+        if (c.getActualMaximum(DAY_OF_YEAR) > 365) c else loop(y + 1)
```
sritchie-stripe: nice fix here, the old version looked wrong.

sritchie-stripe:
Looks great!

PR #299, opened 2016-11-04
remove unnecessary methods in scalacheck 1.14.x

xuwei-k:


non:
👍 despite the transient build failure

PR #301, opened 2016-11-08
Fix retrying code in doPureApply.

### LINKED ISSUE #300: Arbitrary functions fail when the Gen instance on their return type sometimes fails
This means, among other things, that any `Arbitrary` instance that relies on `Gen.suchThat` cannot be used to generate arbitrary functions.

As an example:

```scala
import org.scalacheck._

implicit val nonZero: Arbitrary[Int] = Arbitrary(Arbitrary.arbitrary[Int].suchThat(_ != 0))
val prop = Prop.forAll { (f: String => Int) => f("foo"); true }
```

And then:

```
scala> prop.check
java.lang.StackOverflowError
  ...
```

See alexarchambault/scalacheck-shapeless#50 for a far more insidious example.

non:
This commit fixes a bug pointed out by @alexarchambault. When
constructing an `A => B` value from `Cogen[A]` and `Gen[B]`, we need
to be able to reliably generate a `B` value given a `Seed`. The
`doPureApply` method was given the ability to retry -- but
unfortunately, it used the same result value (with the same seed)
instead of trying a new one, defeating the retry code.

This commit fixes that problem. It adds tests to ensure that filtered
generators that can also produce real values can be used with
Gen.function1. (If a generator can never produce values it will still
be a problem.)

Fixes #300.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -24,6 +24,20 @@ object GenSpecification extends Properties("Gen") {
     arbitrary[Long] flatMap Seed.apply
   )
 
+  property("pureApply #300") = {
+    def testCase[A](gen: Gen[A]): Prop =
+      sizedProp { sz =>
+        val g = Gen.function1(gen)(Cogen[Int])
+        if (sz > 0) forAll(g) { f => f(999); true } else Prop(true)
+      }
+    val p0 = testCase(arbitrary[Int].suchThat(_ != 0))
+    val p1 = testCase(arbitrary[String].suchThat(_ != ""))
+    val p2 = testCase(arbitrary[Boolean].suchThat(_ != false))
+    val p3 = testCase(arbitrary[List[Double]].suchThat(_ != Nil))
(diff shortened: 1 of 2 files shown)


non:
Review by @alexarchambault to confirm that this will improve the situation for Scalacheck-Shapeless.

non:
@rickynils Do you want me to make another PR for this that targets the 1.13.x branch?

alexarchambault:
@non The issue is less a problem with https://github.com/alexarchambault/scalacheck-shapeless/pull/51. I confirm this PR is fine with scalacheck-shapeless.

(Sorry for the late answer. I was taking a plane, then, hem... followed the news...)

sellout:
It _looks_ like this fix isn’t yet released. Could 1.13.5 be published? Because this fix seems critical for slamdata/matryoshka to be able to upgrade to scalacheck-1.13. (It has generic recursive generators, but since they’re generic, it doesn’t know how to explicitly generate a leaf node for any given type, so I _think_ has to use `Gen.fail`, but am open to other approaches).

sellout:
I tried using a 1.14.0 SNAPSHOT with this fix, which worked, but downstream libraries are having binary compatibility issues. Could this be published in a 1.13.x release? (I see @non offered to do that back in Nov.)

PR #307, opened 2016-12-15
Make Gen.identifier properly respect current size

lunaryorn:
When generating the tail of the string reduce the size by one to account for the first character of the generated string.  Otherwise it might generate a string that's one character longer than the current generator size.

Got bitten by this when generating values for an external service that limits the size of certain parameters.

I'm not sure whether this change is entirely correct.  What if the current size is zero?  This change would pass a negative number to the subgenerator.  But can that even happen?

### DIFF
--- a/src/main/scala/org/scalacheck/Gen.scala
+++ b/src/main/scala/org/scalacheck/Gen.scala
@@ -721,10 +721,12 @@ object Gen extends GenArities{
 
   /** Generates a string that starts with a lower-case alpha character,
    *  and only contains alphanumerical characters */
-  def identifier: Gen[String] = (for {
-    c <- alphaLowerChar
-    cs <- listOf(alphaNumChar)
-  } yield (c::cs).mkString)
+  def identifier: Gen[String] = Gen.sized { size =>
+    for {
+      c <- alphaLowerChar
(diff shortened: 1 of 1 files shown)


non:
Hey @lunaryorn sorry for the slow feedback. This looks legit to me, I think the build failure was just transient.

non:
@rickynils The failing test seems to be an anomaly (timeout) and the PR looks good to me, WDYT?

lunaryorn:
@non Never mind, this change is not too important after all :blush:  Thanks for taking a look <3

non:
Do you mean that this PR should no longer be merged, or just that it's low priority?

Looking deeper, it seems like `listOf` is already using `sized`. So maybe we don't need to merge this? If not, feel free to close it.

lunaryorn:
It's low priority for me, since it's easy to work around—just resize the generator outside.

listOf is already size but prepending the extra character means that the final length of the generated is off by one, i.e. size + 1.  To me that was quite unexpected so I opened the PR to change it ☺️ But it's not upon me to decide whether that's expected behaviour 😊

lunaryorn:
Hey 👋 I’d like to close this PR.  I don’t really care for this change anymore, and I’d like to clean up my pull requests page on Github 😊

PR #308, opened 2016-12-19
Enable more compiler warnings, apply respective fixes

sirthias:
In the process of (temporarily) internalizing a sub-set of scalacheck into my project in order to be able to the use the current 1.14 branch I discovered that the current sources produce quite a few compiler warnings that can easily be fixed.

This PR takes these changes upstream.

Most changes don't affect the public API, but in one case they do:
In accordance with `-Ywarn-nullary-unit` the formerly nullary `Test.check` method has now received an empty parameter list.
Since the public API will be broken anyway with 1.14 this change should be ok.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,7 @@ lazy val isRelease = false
 lazy val travisCommit = Option(System.getenv().get("TRAVIS_COMMIT"))
 
 lazy val scalaVersionSettings = Seq(
-  scalaVersion := "2.12.0",
+  scalaVersion := "2.12.1",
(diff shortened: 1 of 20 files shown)


analytically:
Looks like the builds timed out during testing.

non:
Thanks @sirthias!

non:
Hey @sirthias, I just did some work to stabilize our Travis build. I'm going to branch off your PR here and push it and see if I can get it passing Travis. Sorry for the long silence and thanks for your work.

PR #312, opened 2017-01-05
Create ascii generators

agustafson:
Scalacheck is really great, but it would be nice if I could generate more realistic strings. So I've created 2 new generators for Char and String:
* one for printable characters only
* another for all ASCII characters with weighting on printable characters
Let me know if they seem suitable or need some alteration or better documentation.


non:
This looks good to me. 👍 

What do you think @rickynils? Again, we've got Travis flakiness keeping the tests failing.

PR #316, opened 2017-02-27
Fix retrying code in doPureApply.

### LINKED ISSUE #300: Arbitrary functions fail when the Gen instance on their return type sometimes fails

sellout:
(Backport of #301.)

This commit fixes a bug pointed out by @alexarchambault. When
constructing an `A => B` value from `Cogen[A]` and `Gen[B]`, we need
to be able to reliably generate a `B` value given a `Seed`. The
`doPureApply` method was given the ability to retry -- but
unfortunately, it used the same result value (with the same seed)
instead of trying a new one, defeating the retry code.

This commit fixes that problem. It adds tests to ensure that filtered
generators that can also produce real values can be used with
Gen.function1. (If a generator can never produce values it will still
be a problem.)

Fixes #300.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -24,6 +24,20 @@ object GenSpecification extends Properties("Gen") {
     arbitrary[Long] flatMap Seed.apply
   )
 
+  property("pureApply #300") = {
+    def testCase[A](gen: Gen[A]): Prop =
+      sizedProp { sz =>
+        val g = Gen.function1(gen)(Cogen[Int])
+        if (sz > 0) forAll(g) { f => f(999); true } else Prop(true)
+      }
+    val p0 = testCase(arbitrary[Int].suchThat(_ != 0))
+    val p1 = testCase(arbitrary[String].suchThat(_ != ""))
+    val p2 = testCase(arbitrary[Boolean].suchThat(_ != false))
+    val p3 = testCase(arbitrary[List[Double]].suchThat(_ != Nil))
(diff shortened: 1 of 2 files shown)


sellout:
So, 1. that binary compatibility check is really cool, but 2. how did this change break it? 😢

xuwei-k:
I think we should update mima to latest version (0.1.14). Mima 0.1.9 does not support Scala 2.12 and there are some bugs.

- https://github.com/rickynils/scalacheck/blob/1.13.x/project/plugin.sbt#L1
- https://github.com/typesafehub/migration-manager/releases

xuwei-k:
There are some problems :anguished: https://github.com/rickynils/scalacheck/pull/319 https://github.com/rickynils/scalacheck/issues/318

rickynils:
@sellout Can you rebase this PR against `1.13.x`? The binary compatibility issue should hopefully have been fixed now (#318)

sellout:
Rebased and looks good. Thanks.

sellout:
@rickynils Thanks – any chance of a 1.13.6 release with this in it? It’ll allow us to get off a 1.14 SNAPSHOT that makes dealing with conflicting dependencies a pain.

rabbitonweb:
@rickynils could we please release 1.13.6 ?

PR #323, opened 2017-03-19
Add support for atLeastOne method on a generator

ravwojdyla:


ravwojdyla:
Not sure if the travis error is related to the change (?)

non:
I think the Travis flakiness predates this PR (it would be good to figure out how to fix it).

This seems good to me. 👍 

What do you think @rickynils?

PR #324, opened 2017-03-28
Do less work on JS under Travis.

non:
We've been getting build timeouts -- this might help.

A more extreme thing would be to reduce the size ranges under JS,
which I'll do here if this doesn't seem to help.

### DIFF
--- a/src/test/scala/org/scalacheck/PropSpecification.scala
+++ b/src/test/scala/org/scalacheck/PropSpecification.scala
@@ -205,7 +205,7 @@ object PropSpecification extends Properties("Prop") {
       val p0 = Prop.forAll(g1, g2, g3) { (x, y, z) => x == y && y == z }
       val p = p0.useSeed("some name", rng.Seed(n))
       val x = p(params).success
-      val set = (1 to 20).map(_ => p(params).success).toSet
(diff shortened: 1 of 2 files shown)


non:
I think these will help. I've also been considering turning on caching in Travis with something like this:

```
cache:
  directories:
  - $HOME/.sbt/0.13/dependency
  - $HOME/.sbt/boot/scala*
  - $HOME/.sbt/launchers
  - $HOME/.ivy2/cache

before_cache:
  - du -h -d 1 $HOME/.ivy2/cache
  - du -h -d 2 $HOME/.sbt/
  - find $HOME/.sbt -name "*.lock" -type f -delete
  - find $HOME/.ivy2/cache -name "ivydata-*.properties" -type f -delete
```

This won't directly affect the speed of the JS build itself, but will make the overall Travis experience a bit snappier. What do you think @rickynils?

non:
It seems like these are helping. I suggest merging this ASAP (to make it easier for contributors to have passing PRs) and then consider Travis caching in a future PR (since it may require changing the Travis settings, which I don't have permission to do).

rickynils:
Thanks! Really grateful for your contributions. Hope this can solve some of the Travis situation!

PR #325, opened 2017-03-28
Sirthias no warnings

non:
This is a port of @sirthias' PR with *master* merged, to see if Travis will build it cooperatively.

If it passes, then that means we can either merge this PR, or let @sirthias merge *master* to his branch and rebuild. Either way works for me -- I just want to get all these warnings fixed! ;)

Thanks again to Matthias for these fixes.


non:
OK -- it turns out there were Scaladoc problems that I had to fix in addition to merging master.

sirthias:
Thanks, Erik, for taking this on!

PR #326, opened 2017-04-14
fix broken packageBin task

xuwei-k:
http://www.scala-sbt.org/0.13/docs/Custom-Settings.html#Execution+semantics+of+tasks


xuwei-k:
humm maybe still broken

SethTisue:
where in ScalaCheck's release process is `packageBin` invoked? fwiw, I think that's the appropriate location to put a workaround in, given my remarks at https://github.com/rickynils/scalacheck/issues/318#issuecomment-286602193 about circularity

PR #345, opened 2017-07-14
Added support to build on scala.js 1.0.0-M1

cquiroz:
This is a small change that would default to build with scala.js 0.6.18 but would also easily allow for building with 1.0.0-M1. Id like to get scalacheck published up to 1.0.0-M1 to let other projects that depend on scalacheck also build on 1.0.0-M1, e.g. scalatest


rickynils:
Superseded by #348. Also, I'm not sure about injecting the scalajs version through the environment. Maybe it is a good idea, but I have to think about it.

cquiroz:
The use of variable is suggested by scala.js at 
https://www.scala-js.org/news/2017/07/03/announcing-scalajs-1.0.0-M1/

as the way to cross compile across scala.js major versions

It would be useful to get scalacheck on 1.0.0-M1 to port scalatest among other things, see 
https://github.com/scalatest/scalatest/pull/1182

andreaTP:
Actually #348 does not supersede this PR, I do not understand why it has been closed and not actioned.
@rickynils could you please take in consideration to reopen and accept it? Thanks in advance

About the use of Env variables it is suggested here: https://www.scala-js.org/news/2017/07/03/announcing-scalajs-1.0.0-M1/

If it is blocking scalatest to be ported is an important issue for all Scala.Js ecosystem cc. @sjrd

rickynils:
@cquiroz Can you rebase this PR so I can merge it?

cquiroz:
Rebased. When you publish please remember to publish for both scala.js versions

andreaTP:
thanks a lot @rickynils and @cquiroz

PR #357, opened 2017-09-21
Drop Ordering bound on the values in cogenMap

### LINKED ISSUE #356: Cogen.cogenMap should not require ordering on the values
Keys in a `Map` are always distinct. Therefore sorting by the keys should be sufficient:
```scala
implicit def cogenMap[K: Cogen: Ordering, V: Cogen]: Cogen[Map[K, V]] =
  Cogen.it(_.toVector.sortBy(_._1).iterator)
```

joroKr21:
It's safe to sort only by the keys, because they are distinct.

Fixes #356

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ target
 
 /examples/simple-sbt/project/
 /examples/simple-sbt/target/
+
+# Intellij
+.idea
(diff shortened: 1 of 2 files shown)


non:
Looks good to me. 👍

PR #359, opened 2017-10-03
Add Gen.product and Gen.tailRecM

johnynek:
These are two methods I often want with Gen.

I *think* the tailRecM is correct, but I don't know the internal implementation that well so happy to get any feedback on how to improve it.

cc @non

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -323,11 +323,33 @@ object GenSpecification extends Properties("Gen") {
     }
   }
 
+  property("product") = forAll { (a: Int, b: Int) =>
+    forAll(Gen.product(a, b)) {
+      case (a1, b1) => a1 == a && b1 == b
+    }
+  }
+
   property("some") = forAll { n: Int =>
     forAll(some(n)) {
       case Some(m) if m == n => true
       case _ => false
     }
+  }
+
+  property("tailRecM") = forAll { (init: Int) =>
+    val g: ((Int, Int)) => Gen[Either[(Int, Int), Int]] =
+      {
+        case (c, x) if c <= 0 => Right(x)
+        case (c, x) => Left((c - 1, x * Int.MaxValue))
+      }
+
+    val g1 = Gen.tailRecM((10, init))(g)
+    def g2(x: (Int, Int)): Gen[Int] = g(x).flatMap {
+      case Left(y) => g2(y)
+      case Right(x) => Gen.const(x)
+    }
+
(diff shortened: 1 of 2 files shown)


johnynek:
okay, I am much more confident that this is correct now.

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -467,6 +467,53 @@ object Gen extends GenArities{
...
+    // This is the "Reader-style" appoach to making a stack-safe loop:
+    // we put one outer closure around an explicitly tailrec loop
+    gen[B] { (p: P, seed: Seed) =>
+      tailRecMR(a0, seed, Set.empty) { (a, seed) => fn(a).doApply(p, seed) }
+    }
+  }
+
+  /** Build a pair from two generators
+   */
+  def product[A, B](ga: Gen[A], gb: Gen[B]): Gen[(A, B)] =
```
ceedubs: The implementation is a bit different, but there is a [zip](https://github.com/rickynils/scalacheck/blob/35d5948a9185f509247b165ae9f02ba8c105c008/src/main/scala/org/scalacheck/GenArities.scala#L270) method. Is this intended to be available in addition to that?

johnynek: honestly, I didn't know about it. I think this implementation is more efficient but I am happy to remove it (or change the zip 2 implementation to use this approach).

My main interest in the tailRecM.

ceedubs: @johnynek just to provide some context for that last comment: I'm not a Scalacheck maintainer; I just happened to see this change and I was pretty sure that a method similar to `product` existed, so I thought that I would point it out. I have no opinions on how to proceed regarding `product`, but I certainly could see benefit in `tailRecM` existing.

non: Let's remove `product` here and devote the energy to porting that strategy over to the generated `zip` implementations. I agree with @johnynek that this approach is probably more efficient, so let's definitely do that.

jvm/src/test/scala/org/scalacheck/GenSpecification.scala
```diff
@@ -323,13 +323,52 @@ object GenSpecification extends Properties("Gen") {
...
+
+    val finalG2 = g2((10, init))
+
+
+    val params = Gen.Parameters.default
+
+    seeds.forall { seed =>
+      g1.pureApply(params, seed) == finalG2.pureApply(params, seed)
+    }
+  }
```
non: This is a great property!

non:
Once we remove `product` I'm +1 on merging this.

non:
👍

PR #363, opened 2017-10-18
add proxied generator to access Parameters and Seed within a Gen

ssanj:
This generator will pass the Parameters and Seed supplied from the outer generator to the inner/proxied one. I needed to access the framework generated seed in one of the generators I was building it would have been useful to have this function. I'm not sure if it is really necessarily useful for anyone else. Happy to hear thoughts on this even if this PR is not merged.


ashawley:
Shouldn't `withPerturb` allow you to access the generated seed?

ssanj:
@ashawley you're right! Not sure how I missed it! Thanks!

PR #364, opened 2017-10-20
Fixes the type parameter in a call to arbitrary in a property

mdimjasevic:
This was probably a copy-paste error when writing properties so I believe there is no need for much discussion here.


mdimjasevic:
@rickynils, do you have a moment to take a look at this pull request, which should be trivial?

mdimjasevic:
Can someone take a look at this simple modification and hopefully merge it?

rickynils:
Thanks! And sorry for being so slow...

Issue #89, opened 2014-02-20
mapOfN will not work well if there is a high probability of duplicate keys


alexeyr:
`mapOfN` generates exactly `n` key-value pairs, constructs a map from them, and checks that the size is `n`. This will fail and be discarded if there are any duplicate keys. Given that, e.g. numeric generators produce specific values with high probability, this is quite likely to happen.

gkazior:
I got similar problem with:

```scala
Gen.containerOfN[Set, String](2, Gen.oneOf("one", "two"))
```
which usually returns Set(one, two) but sometimes Set() 
Unfortunately my case was much more complicated.

The documentation states: 
"The size of the generated container is given by n."
Which is not true!

```scala
val gen=Gen.containerOfN[Set, String](2, Gen.oneOf("one", "two")).suchThat(_.size == 2)
gen.sample
```

I understand the problem with:
```scala
Gen.containerOfN[Set, String](3, Gen.oneOf("one", "two"))
```

BTW. I wanted a list with unique elements and given length. Possibly there is a better method.

Great lib!

PR #370, opened 2017-11-14
Prevent shrinking for listOfN

nigredo-tori:
Add `.suchThat(_.size == n)` to `Gen` created by `listOfN` to shield the output from shrinking in case of a failure.

There are a few issues in this repository with `containerOfN` and `buildableOfN` producing containers of incorrect size. They all seem to arrive at the same conclusion: creating containers of required size is difficult in general case. However, if we know the exact container we're using, it can be trivial.

This changes the behavior of `listOfN` to agree with its doc comment: the resulting generator now produces lists of exactly the given length.

### DIFF
--- a/src/main/scala/org/scalacheck/Gen.scala
+++ b/src/main/scala/org/scalacheck/Gen.scala
@@ -650,9 +650,10 @@ object Gen extends GenArities{
    *  `nonEmptyContainerOf[List,T](g)`. */
   def nonEmptyListOf[T](g: => Gen[T]) = nonEmptyBuildableOf[List[T],T](g)
 
-  /** Generates a list of the given length. This method is equal to calling
-   *  `containerOfN[List,T](n,g)`. */
-  def listOfN[T](n: Int, g: Gen[T]) = buildableOfN[List[T],T](n,g)
+  /** Generates a list of the given length. This method is equivalent to calling
+   *  `containerOfN[List,T](n,g).suchThat(_.size == n)`. */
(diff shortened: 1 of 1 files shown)


ashawley:
FWIW, there's a comment about not being able to impose a size and refers to #89.

You can use `forAllNoShrink` to prevent shrinking:

```
    val g = Gen.listOfN(3, Gen.const(1))
    Prop.forAllNoShrink(g) { threeOnes: List[Int] =>
      threeOnes.length == 4
    }
```

Gives the following failure:

```
[info] ! PropSpecification.listOf: Falsified after 0 passed tests.
[info] > ARG_0: List(1, 1, 1)
```

Rather than:

```
[info] ! PropSpecification.listOf: Falsified after 0 passed tests.
[info] > ARG_0: List()
[info] > ARG_0_ORIGINAL: List(1, 1, 1)
```

Here's the doc for `buildableOfN`:

> Generates a container of any `Traversable` type for which there exists an implicit `org.scalacheck.util.Buildable` instance. The elements in the container will be generated by the given generator. The size of the generated container is limited by `n`. Depending on what kind of container that is generated, the resulting container may contain fewer elements than `n`, but not more. If the given generator fails generating a value, the complete container generator will also fail.

rickynils:
@nigredo-tori See @ashawley's workaround. Introducing `suchThat`-calls inside the ScalaCheck API currently breaks more things than it fixes, due to ScalaCheck's slightly broken implementation with regards to co-generators.

PR #373, opened 2017-11-27
Fix import statements

ashawley:
There a few cases in the code where a wildcard import statement is
used unnecessarily:

    import org.scalacheck._

Which could be avoided by just using multiple package declarations.

For example,

    package org.scalacheck
    package util

Instead of 

    package org.scalacheck.util

This change **in total** is more deletions than additions in lines of codes, and should avoid people copy-pasting this anti-pattern any further in the code tree.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/commands/CommandsSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/commands/CommandsSpecification.scala
@@ -7,9 +7,8 @@
 **  There is NO WARRANTY. See the file LICENSE for the full text.          **
 \*------------------------------------------------------------------------ */
 
-package org.scalacheck.commands
-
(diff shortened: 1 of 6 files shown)


rickynils:
Your PR is more deletions than additions because some white lines are removed.

I'm probably missing something, but why is the first example worse than the second, below?

```
package org.scalacheck.commands
import org.scalacheck._
```

```
package org.scalacheck
package commands
```

In my eyes, the exact same things will be brought into scope. I prefer the first way slightly, because you immediately see what package we're in, and don't have to concatenate `org.scalacheck` and `commands` in your head.

ashawley:
Yes, package-import vs. package-package are the same number of lines sans whitespace.  I just don't know of any other projects doing it this way.

PR #381, opened 2018-01-18
Fix Shrink code sample in UserGuide

kantlove:
The code sample in section [Test Case Minimisation](https://github.com/rickynils/scalacheck/blob/master/doc/UserGuide.md#test-case-minimisation) contains calls to `shrink` method without instances of `Shrink`
```scala
/** Shrink instance of 2-tuple */
implicit def shrinkTuple2[T1,T2](implicit s1: Shrink[T1], s2: Shrink[T2]
): Shrink[(T1,T2)] = Shrink { case (t1,t2) =>
  (for(x1 <- shrink(t1)) yield (x1, t2)) append
  (for(x2 <- shrink(t2)) yield (t1, x2))
}
```
I add the instances like this
```scala
  (for(x1 <- s1.shrink(t1)) yield (x1, t2)) append
  (for(x2 <- s2.shrink(t2)) yield (t1, x2))
```

### DIFF
--- a/doc/UserGuide.md
+++ b/doc/UserGuide.md
@@ -786,8 +786,8 @@ is defined in ScalaCheck:
 /** Shrink instance of 2-tuple */
 implicit def shrinkTuple2[T1,T2](implicit s1: Shrink[T1], s2: Shrink[T2]
 ): Shrink[(T1,T2)] = Shrink { case (t1,t2) =>
-  (for(x1 <- shrink(t1)) yield (x1, t2)) append
-  (for(x2 <- shrink(t2)) yield (t1, x2))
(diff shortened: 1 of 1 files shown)


rickynils:
Well, the User Guide mentions that the code is a snippet from `Shrink.scala` (https://github.com/rickynils/scalacheck/blob/master/src/main/scala/org/scalacheck/Shrink.scala#L113). The `shrink` method comes from the scope inside `Shrink`. This PR changes the code so that is no longer a snippet from `Shrink.scala`.

A less obtrusive change could maybe be to add `import org.scalacheck.Shrink.{shrink}`.

kantlove:
@rickynils Oh I see. In that case, I think this change is not necessary.

PR #383, opened 2018-02-01
Scala 2.13.0-M3

xuwei-k:


ashawley:
scala-js in `.travis.yml` needs to be 1.0.0-M3, not 1.0.0-M2

ashawley:
And then 1.0.0-M3 needs to be excluded from being run on 2.10.6 on Travis, since 2.10 support was dropped in 1.0.0-M3

https://www.scala-js.org/news/2018/02/01/announcing-scalajs-1.0.0-M3/

PR #391, opened 2018-03-29
[WIP] Add support for Scala 2.13-M4

julienrf:
In the process of making the new collections as compatible as possible with the old collections, I’ve migrated this project to find out where the incompatibilities are.

Most of the changes shown in the diff have been automatically applied by a scalafix [rewrite rule](https://github.com/julienrf/collection-strawman/blob/07dd6bfe272237fece81d50dfb66e996c5e52270/scalafix/2.13/rules/src/main/scala/fix/Collectionstrawman_v0.scala). In some cases I had to take decisions or just rewrite things differently. I’m using the [scala-collection-compat](http://github.com/scala/scala-collection-compat) library to make 2.13’s API available on 2.12 and prior Scala versions. I use two specific source directories, for 2.13 on one side, and for 2.12 and prior Scala versions on the other side, to implement a custom collection (`ArrayListBuilder`).

The changes automatically handled by the rewrite rule are the following:

- replace `Traversable` with `Iterable`,
- add parens to calls to `iterator()` and `newBuilder()`,
- replace `Stream#append` with `Stream#lazyAppendAll`.

The areas that are *not* handled by the rewrite rule are the following:

- `CanBuildFrom` usage (here, replaced with `Factory`),
- `scala.Seq` usage (here, I chose the conservative approach that consists in using `scala.collection.Seq`, which is the same on 2.12 and 2.13, instead of `scala.Seq`, which expands to `immutable.Seq` on 2.13),
- implementation of a type provided by the collection API (here, `ArrayListBuilder`).


julienrf:
We can currently only `jvm/test` on 2.12 and 2.13, Scala.js and native aren’t supported yet.

julienrf:
@rickynils It would be useful for us to have scalacheck release compatible with Scala 2.13.0-M4-pre-20d3c21. Would it be possible for you to publish such a release?

rickynils:
Superseded by #411

PR #392, opened 2018-03-29
make Arbitrary[String] return an arbitrary string

martijnhoekstra:
To open discussion to broaden what an arbitrary string entails.

ScalaCheck is so useful to me because it often finds edge-cases that I didn't. To make `Arbitrary[String]` by default return a relatively well-behaved string runs counter to that.

In reality, Strings are messy. They can contain private use characters, supplemental characters, unassigned characters, and even invalid unicode with unpaired surrogate characters.

If a program thinks it can deal with arbitrary strings, and wants to check that with scalacheck, right now it avoids checking the spots where the bugs probably are.

Dealing with some subset of strings is entirely reasonable for some application, and this PR adds some generators for some different definitions of valid strings.

### DIFF
--- a/src/main/scala/org/scalacheck/Arbitrary.scala
+++ b/src/main/scala/org/scalacheck/Arbitrary.scala
@@ -150,7 +150,7 @@ private[scalacheck] sealed trait ArbitraryLowPriority {
 
   /** Arbitrary instance of String */
   implicit lazy val arbString: Arbitrary[String] =
-    Arbitrary(arbitrary[List[Char]] map (_.mkString))
+    Arbitrary(Gen.anyStr)
 
   /** Arbitrary instance of Date */
   implicit lazy val arbDate: Arbitrary[java.util.Date] =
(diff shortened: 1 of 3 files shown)


martijnhoekstra:
Note that this PR will probably break peoples tests. Which fraction of those test breakages are actual bugs is more difficult to answer.

dwijnand:
should the same be done on `arbChar`?

martijnhoekstra:
@dwijnand IMO yes, for the same reasons. But let's see what rickynils wants for his library. I suspect there will be some discussion both on the concept and the implementation.

morgen-peschke:
I've recently needed something similar and I found that just having invalid input wasn't enough to verify the properties I wanted, I also needed to know where in the string the invalid characters appeared.

Because of this I ended up writing a generator which produced arbitrary sequences of invalid `Byte`s that I could splice into a `Array[Byte]`, which turned out to be more useful.

This approach would also avoid breaking current tests.

larsrh:
Closing this one for inactivity. If there's more interest in this, please reopen.

PR #393, opened 2018-04-05
Prefer immutable.Seq and Iterable

ashawley:
In preparation for the vaunted 2.13 collections library changes, I've cherry-picked a few of Julien Richard-Foy's changes in his 2.13.0-M4 porting work in #391.    In the interest of better understanding the impending 2.13 release changes, it would be good to at least make piecewise progress.  

Julien was looking at using the scala-collections-compat library and preserving the internal uses of `mutable.Seq`.  I investigated just switching to using `immutable.Seq` and making the switch from `Traversable` to `Iterable`.

These changes don't seem very onerous and they seem to work across multiple Scala versions and targets (including js and native).


ashawley:
There was a strange issue where `Array` and `mutable.ArrayBuffer` wouldn't implicitly convert to`immutable.Seq`.  It normally should.

```scala
def f[T](a: collection.mutable.ArrayBuffer[T]):collection.mutable.Seq[T] = {
  if (a.isEmpty) collection.mutable.Seq.empty[T]
  else collection.mutable.ArrayBuffer.empty[T].to[collection.mutable.Seq]
}
f: [T](a: collection.mutable.ArrayBuffer[T])collection.mutable.Seq[T]
```

Seemingly, this was because `org.scalacheck.util.SerializableCanBuildFroms` had implicits that interfered.

src/main/scala/org/scalacheck/Test.scala
```diff
@@ -11,6 +11,8 @@ package org.scalacheck
 
 import Prop.Arg
 
+import scala.collection.immutable.Seq
+
```
ashawley: This is because `Test.checkProperties` return type is defined as `Seq[(String,Result)]`.

ashawley:
I didn't realize MiMa was disabled until #399 put it back.  

Turns out there are a lot of compatibility changes if `immutable.Seq` is used:

```
sbt> jvm/mimaReportBinaryIssues
[info] scalacheck: found 7 potential binary incompatibilities while checking
[info] against org.scalacheck:scalacheck_2.12:1.14.0 
```

Issue #352, opened 2017-08-10
Generating a list of size N where each element is distinct.


wjlow:
I'm proposing something like:

```
def sizedDistinctListOf[T](size: Int, gen: Gen[T], eq: (T, T) => Boolean): Gen[List[T]]
```

In my own codebase, I've got something similar where `T` is constrained by the `Eq` typeclass from Cats. This allows me to define what distinctiveness means. From my understanding, ScalaCheck has no dependency on Cats.

For instance, you may have a `case class Person(name: String, age: Int)` and for a specific scenario, you might want to generate `List[Person]` where each element contains a different `name` and `age` pair, whereas for some other cases, you might want to generate ` List[Person]` where each element has a different `age`.

Happy to submit a PR if people find this useful.

ashawley:
How about?

```scala
val genPersonSameName: Gen[Person] = arbitrary[Person].map(_.copy(name = "Same Name"))
val genPersonSameAge: Gen[Person]  = arbitrary[Person].map(_.copy(age  = 42))
```

Then you can generate distinct lists with `Set` and `containerOfN`:

```scala
val genPersonDiffName: Gen[List[Person]] = containerOfN[Set, Person](n, genPersonSameAge).map(_.toList)
val genPersonDiffAge: Gen[List[Person]]  = containerOfN[Set, Person](n, genPersonSameName).map(_.toList)
```

I don't enough about cats to know if `Eq` cooperates well with scala collections.

ashawley:
Looks like there is an effort to make cats work with scala collections, including `Set`:

https://github.com/non/alleycats

And also an initiative to build cats extensions to Scalacheck:

https://github.com/non/cats-check

wjlow:
Thanks for the suggestion @ashawley. I feel like this approach isn't as obvious though. Can we provide a simple function that generates a collection of T _provided_ a way to define distinctiveness?

Your approach works and I like how it works by composing Gens. However I feel like it's easier to provide a function that determines equality `(A,A) => Boolean` as opposed to having to write a Gen for for it instead. 

Not sure if this makes sense.

morgen-peschke:
I've found this sort of thing useful in the past, and have opened #394 with a potential implementation.

morgen-peschke:
@ashawley - that's definitely an important point. 

I think the main point of this feature request isn't to try to guarantee a particular size (though a best effort there is always appreciated), as much as it is to have some way of customizing what "distinct" means for a particular data set.

[...]

PR #394, opened 2018-04-06
Add factory for collections of a fixed number of distinct elements

morgen-peschke:
Implements request in #352


non:
Hi @morgen-peschke,

I think it's worth adding something like this, but I have some concerns with this PR:

 - The algorithm is quadratic, we have to take the `.size` every time we add an element, and call `areEqual` for every element in the list so far every time we add an element.
 - It seems like the most common cases here are situations where the caller either wants to project the "uniqueness" down to a field, or use a custom ordering, both of which can be more efficient (O(n) or O(n log n)).
 - I'd prefer to use higher-level combinators to define the method(s) being added.

For example, here's a way to implement your current algorithm using `tailRecM`:

```scala
  // quadratic complexity, so not great as n gets large
  def distinctByFn[T](n: Int, g: Gen[T], maxCollisions: Int = 1000)(p: (T, T) => Boolean): Gen[List[T]] =
    tailRecM((List.empty[T], 0, 0)) { case (ts, sz, col) =>
      if (col >= maxCollisions) Gen.fail
      else if (sz >= n) Gen.const(Right(ts))
      else g.map { t =>
        if (ts.exists(p(_, t))) Left((ts, sz, col + 1))
        else Left((t :: ts, sz + 1, col))
      }
    }
```

I'd prefer to add the following two methods instead:

```scala
def setOfN[T: Ordering](n: Int, g: Gen[T], maxCollisions: Int = 1000): Gen[SortedSet[T]] = ???

def distinctN[T, U](n: Int, g: Gen[T], maxCollisions: Int = 1000)(f: T => U): Gen[List[T]] = ???
```

What do you think? It's true that using `(T, T) => Boolean` is more flexible but I don't see a good way to do it efficiently.

non:
@morgen-peschke On reflection I think having multiple functions is OK, as long as we make sure to highlight the possible performance issues in the Scaladoc.

We're fortunate that none of the methods we're writing are impossible to write from third-party code, so we don't need to cover every possible case (users will be able to write these methods themselves if needed).

How do you feel about implementing these three?

```scala
def setOfN[T](n: Int, g: Gen[T], maxCollisions: Int = 1000): Gen[Set[T]]

def distinctN[T, U](n: Int, g: Gen[T], maxCollisions: Int = 1000)(distinctBy: T => U): Gen[List[T]]

def distinctNByFn[T, U](n: Int, g: Gen[T], maxCollisions: Int = 1000)(areEquiv: (T, T) => Boolean): Gen[Set[T]]
```

I think the first will be the most common: people want a set of N different things, using the normal rules for equality/hashing, and without an `Ordering[T]` instance needed. I decided that maybe imposing an `Ordering` constraint (to allow things like projection) is worse than just having `distinctN` support projection directly (again, using hashing/equality). The third corresponds to your method: as you say it's the most general but also has the worst performance, so people will presumably only reach for it when the first two don't work (and hopefully for relatively small N).

How does this sound?

(I left out the `Buildable` support here but I'm fine with including these variants, I just didn't feel like writing out all the types in this comment! 😉)

non:
Oh, I forgot to mention why I prefer `maxCollisions`: I think it makes it easier for users to impose an upper-bound on the number of times a generator will be run. With the previous approach, `N * max` is the number of generations to be attempted, whereas now it's `N + max`. If someone misunderstand the new approach, they'll see generator failures and potentially raise the threshold, whereas if they get it wrong in the other case their generators might just get really slow without them realizing why.

Either way the author will probably need to do some statistics if they think they'll reach a high likelihood of failure and want to give their generator a good chance of succeeding (e.g. generating 255 unique `Byte` values or something).

morgen-peschke:
Looks good to me. I'll probably get it in sometime this weekend.

> (I left out the Buildable support here but I'm fine with including these variants, I just didn't feel like writing out all the types in this comment! 😉)

I feel you. I really hope we end up with nicer syntax for this stuff in dotty :)

non:
Great, thanks! Looking forward to this, it's a good feature to add.

morgen-peschke:
@non, how's it look?

bchallenor:
I independently implemented `setOfN` today (and even called it that) so I would love to see this in scalacheck proper.

morgen-peschke:
@non, would you like me to close this, or keep up the current cadence of a ping every couple of months?

morgen-peschke:
ping

morgen-peschke:
@rickynils, I thought @non was doing the review on this one but haven't heard from him for a while, is there someone else that can pick this up?

morgen-peschke:
ping @rickynils or @non

SethTisue:
@morgen-peschke are you interested in rebasing this? if so, I think we could proceed with review (and presumably merge) without Erik — his feedback before was pretty thorough, so if you think you've complied with it, then perhaps we don't need a second deep dive from him and the rest of us could sign off instead

morgen-peschke:
@SethTisue I'm interested, the blocking resource is time. 

It's been almost 2 years since I've touched this code, so I'd have to spend time getting context again that I simply cannot spare at the moment, and it'll probably be a couple months before I can carve out enough free time to get back to this 😞

SethTisue:
okay — I'm going to close this then, but maybe someone will see the notifications and decide to volunteer to pick it up

Issue #230, opened 2016-04-07
Compilation errors with Scala 1.12.0-M4


rickynils:
Compiling ScalaCheck 1.13.0 with Scala 1.12.0-M4 results in a lot of errors like these:

```
[error] src/main/scala/org/scalacheck/Cogen.scala:25: missing parameter type
[error]     Cogen((seed, s) => perturb(seed, f(s)))
[error]            ^
[error] src/main/scala/org/scalacheck/Cogen.scala:25: missing parameter type
[error]     Cogen((seed, s) => perturb(seed, f(s)))
```

These errors are introduced by M4, compilation with M3 works fine. Additionally, M4 works fine with ScalaCheck 1.11 and 1.12.

dwijnand:
Ouch. The 2.12.x community build has scalacheck pinned at 1.11.6: https://github.com/scala/community-builds/blame/2.12.x/common.conf#L66

xuwei-k:
`Cogen` is SAM (single abstract method) type, and there are some overload methods.

```scala
def apply[T](implicit ev: Cogen[T]): Cogen[T] = ev

def apply[T](f: (Seed, T) => Seed): Cogen[T] = 
```

https://github.com/rickynils/scalacheck/blob/1.13.0/src/main/scala/org/scalacheck/Cogen.scala

xuwei-k:
similar discussion? https://github.com/scala/scala/pull/4971#issuecomment-202977627

SethTisue:
attn @adriaanm, here's another example of overloading trouble involving SAM support

adriaanm:
That's a fun one! The smallest source-compatible hack I can think of is `def apply[T](implicit ev1: Cogen[T], ev2: Cogen[T]): Cogen[T] = ev1`... Don't tell anyone I recommended that, though :sunglasses: 

Sadly, I don't see how we can improve overload resolution to fix this without pushing it over the complexity brink that it's teetering.

rickynils:
@adriaanm So ScalaCheck follows its tradition of torturing scalac... The workaround works, but I guess I must either bump the ScalaCheck version or create a separate branch for M4.

sjrd:
You can make it both source-compatible and binary compatible like this:
```scala
// binary compatible
private[enclosingpackage] def apply[T](ev: Cogen[T]): Cogen[T] = ev
// source compatible
def apply[T](implicit ev: Cogen[T], dummy: DummyImplicit): Cogen[T] = ev
```

[...]

PR #395, opened 2018-04-21
Revert 2.12.0-M4 changes

ashawley:
These were changes that were made for fixing errors in a 2.12 milestone and also a fix that avoided breaking binary compatability in 1.13.0 for the bootstrap process.

This reverts #230 it was already partially reverted in 8abd189

### DIFF
--- a/src/main/scala/org/scalacheck/Cogen.scala
+++ b/src/main/scala/org/scalacheck/Cogen.scala
@@ -30,8 +30,7 @@ sealed trait Cogen[T] extends Serializable {
 
 object Cogen extends CogenArities with CogenLowPriority {
 
-  // See https://github.com/rickynils/scalacheck/issues/230 for dummy expl.
-  def apply[T](implicit ev: Cogen[T], dummy: Cogen[T]): Cogen[T] = ev
+  def apply[T](implicit ev: Cogen[T]): Cogen[T] = ev
 
   def apply[T](f: T => Long): Cogen[T] = new Cogen[T] {
     def perturb(seed: Seed, t: T): Seed = seed.reseed(f(t))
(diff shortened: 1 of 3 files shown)


xuwei-k:
This change cause ambiguous error with Scala 2.11.12 and `-Xexperimental` option 😕 

```
scala> org.scalacheck.Cogen[Int]{(a: Seed, b: Int) => a}
<console>:16: error: ambiguous reference to overloaded definition,
both method apply in object Cogen of type (f: (org.scalacheck.rng.Seed, Int) => org.scalacheck.rng.Seed)org.scalacheck.Cogen[Int]
and  method apply in object Cogen of type (implicit ev: org.scalacheck.Cogen[Int])org.scalacheck.Cogen[Int]
match argument types ((org.scalacheck.rng.Seed, Int) => org.scalacheck.rng.Seed)
       org.scalacheck.Cogen[Int]{(a: Seed, b: Int) => a}
                           ^
```

src/main/scala/org/scalacheck/Cogen.scala
```diff
@@ -30,8 +30,7 @@ sealed trait Cogen[T] extends Serializable {
 
 object Cogen extends CogenArities with CogenLowPriority {
 
-  // See https://github.com/rickynils/scalacheck/issues/230 for dummy expl.
-  def apply[T](implicit ev: Cogen[T], dummy: Cogen[T]): Cogen[T] = ev
```
fommil: this broke bincompat

ashawley: Yes it did.  I believe it should have only been released in 1.14.0, and not in 1.13.x.

fommil: hmm. Interesting. I am seeing the breakage in 1.13.x.

fommil: it is possible something is transitively pulling in 1.14.x in my project. I'll check that.

Issue #397, opened 2018-04-23
compile error in build.sbt


xuwei-k:
since https://github.com/rickynils/scalacheck/commit/2cdfc5512a48672c6c1439f1aca4271f8c05353e

https://github.com/rickynils/scalacheck/blob/2cdfc5512a48672c6c1439f1aca4271f8c05353e/build.sbt#L81-L84

```
/home/travis/build/[secure]/scalacheck/build.sbt:83: error: type mismatch;
 found   : String("org.scalacheck")
 required: ?{def %%%(x$1: ? >: String("scalacheck")): ?}
Note that implicit conversions are not applicable because they are ambiguous:
 both method toScalaJSGroupID in trait DependencyBuilders of type (groupID: String)org.scalajs.sbtplugin.impl.ScalaJSGroupID
 and method toPlatformDepsGroupID in object autoImport of type (groupID: String)org.portablescala.sbtplatformdeps.PlatformDepsGroupID
 are possible conversion functions from String("org.scalacheck") to ?{def %%%(x$1: ? >: String("scalacheck")): ?}
      Set("org.scalacheck" %%% "scalacheck" % "1.14.0")
          ^
/home/travis/build/[secure]/scalacheck/build.sbt:83: error: value %%% is not a member of String
      Set("org.scalacheck" %%% "scalacheck" % "1.14.0")
                           ^
[error] sbt.compiler.EvalException: Type error in expression
```

rickynils:
Fixed in #399

PR #399, opened 2018-04-23
Re-enable mima check

ashawley:
Apparently, MiMa was disabled.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,12 +78,7 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
   // don't use fatal warnings in tests
   scalacOptions in Test ~= (_ filterNot (_ == "-Xfatal-warnings")),
 
-  mimaPreviousArtifacts := (
-    if (CrossVersion isScalaApiCompatible scalaVersion.value)
-      Set("org.scalacheck" %%% "scalacheck" % "1.14.0")
-    else
-      Set.empty
-  ),
+  mimaPreviousArtifacts := Set("org.scalacheck" %% "scalacheck" % "1.14.0"),
 
   publishTo := {
     val nexus = "https://oss.sonatype.org/"
--- a/project/MimaSettings.scala
+++ b/project/MimaSettings.scala
@@ -22,6 +22,7 @@ object MimaSettings {
   )
 
   private def removedPrivateClasses = Seq(
+    "org.scalacheck.Platform$EnableReflectiveInstantiation"
   )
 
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -78,12 +78,7 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
   // don't use fatal warnings in tests
   scalacOptions in Test ~= (_ filterNot (_ == "-Xfatal-warnings")),
 
-  mimaPreviousArtifacts := (
-    if (CrossVersion isScalaApiCompatible scalaVersion.value)
-      Set("org.scalacheck" %%% "scalacheck" % "1.14.0")
-    else
-      Set.empty
-  ),
+  mimaPreviousArtifacts := Set("org.scalacheck" %% "scalacheck" % "1.14.0"),
```
ashawley: MiMa is failing for the Scalajs and Scala native builds in similar ways:

```
[info] scalacheck: found 2 potential binary incompatibilities while checking
[info] against org.scalacheck:scalacheck_2.13.0-M3:1.14.0 
[error]  * class org.scalacheck.Platform#EnableReflectiveInstantiation
[error]    does not have a correspondent in current version 
[error]    filter with:
[error]    ProblemFilters.exclude[MissingClassProblem]("org.scalacheck.Platform$EnableReflectiveInstantiation")
[error]  * method deserializeTask(java.lang.String,scala.Function1)org.scalacheck.ScalaCheckRunner#BaseTask
[error]    in class org.scalacheck.ScalaCheckRunner has a different result
[error]    type in current version, where it is sbt.testing.Task rather than
[error]    org.scalacheck.ScalaCheckRunner#BaseTask 
[error]    filter with:
[error]    ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.scalacheck.ScalaCheckRunner.deserializeTask")
```

I assumed Mima supports those platforms, but maybe it doesn't?  Or is there more sbt magic needed for it to work?

ashawley:
This fixes the build as reported in #397, but could use a review of of the Scalajs and Native bits.

ceedubs:
I'm not a scalacheck maintainer, so this should be taken with a grain of salt, but FWIW I'd advocate erring on the side of merging this sooner rather than later even if we don't receive timely input from a Scala.js/Scala Native expert. Right now the master branch doesn't compile which prevents people from working on any other changes (I have a couple of minor PRs that I want to prepare). To me that seems likely to be a larger issue than a mima exception that we aren't sure about in the non-JVM context.

PR #403, opened 2018-04-27
partially revert #395

xuwei-k:
see https://github.com/rickynils/scalacheck/pull/395#issuecomment-383442015


non:
Thanks!

ashawley:
In which version of ScalaCheck?  I believe all the 1.13.x versions have d25dde5 in them which is when it was added originally.  It was removed for 1.14.0 but reverted after 1.14.0 was released.    So I don't think there was a 1.13 without it.

PR #405, opened 2018-04-30
Fix console scalacOptions

ashawley:
It's not possible to hack on Scalacheck from the REPL since #353.

    sbt> jvm/console
    [info] Starting scala interpreter...

    scala> import org.scalacheck.Prop
    import org.scalacheck.Prop
    <console>:11: warning: Unused import
           import org.scalacheck.Prop
                                 ^
    error: No warnings can be incurred under -Xfatal-warnings.

    scala> Prop.passed
    <console>:12: error: not found: value Prop
           Prop.passed
           ^

On line 75 of `build.sbt`, this line is supposed to disable it, but the filter is no longer working:

```scala
scalacOptions in (Compile, console) ~= {_.filterNot("-Ywarn-unused-import" == _)},
```


non:
Thanks!

PR #407, opened 2018-05-03
Fix compilation on Java 9

### LINKED ISSUE #406: Compiling on Java 9 fails
Just noticed that compiling with Java 9 gives a deprecation error:

```
[error] jvm/src/main/scala/org/scalacheck/Platform.scala:66:44: method newInstance in class Class
[error] is deprecated: see corresponding Javadoc for more information.
[error]     else Class.forName(name, true, loader).newInstance.asInstanceOf[AnyRef]
[error]                                            ^
[error] one error found
[error] (jvm / Compile / compileIncremental) Compilation failed
[error] Total time: 31 s, completed May 3, 2018, 8:52:03 AM
```

Presumably, this is because both `-deprecation` and `-Xfatal-warnings`  is enabled for the compiler.

ashawley:
Fixes #406.

This compiles on Java 9, but it doesn't seem there are tests here to verify that this change works.

Seems this code is related to running tests in sbt.  If ScalaCheck's own test suite runs, it may be proof that it works.  Running the debugger and adding a breakpoint, it does appear to be called, for what that's worth.

### DIFF
--- a/jvm/src/main/scala/org/scalacheck/Platform.scala
+++ b/jvm/src/main/scala/org/scalacheck/Platform.scala
@@ -63,7 +63,7 @@ private[scalacheck] object Platform {
 
   def newInstance(name: String, loader: ClassLoader, paramTypes: Seq[Class[_]])(args: Seq[AnyRef]): AnyRef =
     if(!args.isEmpty) ???
-    else Class.forName(name, true, loader).newInstance.asInstanceOf[AnyRef]
+    else Class.forName(name, true, loader).getDeclaredConstructor().newInstance().asInstanceOf[AnyRef]
 
   def loadModule(name: String, loader: ClassLoader): AnyRef =
     Class.forName(name + "$", true, loader).getField("MODULE$").get(null)


jvm/src/main/scala/org/scalacheck/Platform.scala
```diff
@@ -63,7 +63,7 @@ private[scalacheck] object Platform {
 
   def newInstance(name: String, loader: ClassLoader, paramTypes: Seq[Class[_]])(args: Seq[AnyRef]): AnyRef =
     if(!args.isEmpty) ???
-    else Class.forName(name, true, loader).newInstance.asInstanceOf[AnyRef]
+    else Class.forName(name, true, loader).getDeclaredConstructor(paramTypes:_*).newInstance().asInstanceOf[AnyRef]
```
SethTisue: if `args` is empty, then isn't `paramTypes` empty too?

ashawley: Good point!  Lets try without it.

SethTisue:
the Travis-CI failure seems spurious

ashawley:
Agree.  It was not even a spurious error with ScalaCheck's test suit, this one was Travis having trouble kickstarting the build for Scala native.

PR #408, opened 2018-05-12
Option Arbitrary delegates to Option Gen

### LINKED ISSUE #401: Discrepancy between Gen and Arbitrary for Option
The `Gen` [instance](https://github.com/rickynils/scalacheck/blob/2cdfc5512a48672c6c1439f1aca4271f8c05353e/src/main/scala/org/scalacheck/Gen.scala#L541-L543) for `Option` generates a `Some` 90% of the time (see #286).

The `Arbitrary` [instance](https://github.com/rickynils/scalacheck/blob/2cdfc5512a48672c6c1439f1aca4271f8c05353e/src/main/scala/org/scalacheck/Arbitrary.scala#L337-L345) for `Option` doesn't delegate through to the `Gen` instance but instead uses the `size` as a factor for the frequency of `Some`.

It seems to me like Scalacheck should settle on one of these two approaches for generating `Option`, have the `Gen` use it, and have the `Arbitrary` instance simply delegate to the `Gen` instance. The idea of using the `size` for the frequency of `Some` is an interesting one, though it's not immediately obvious to me whether one approach is better than the other. Also the `Arbitrary` instance is currently resizing the inner `Gen[A]` instance to half of `size` which stumped me a little and [appears to have done the same](https://github.com/rickynils/scalacheck/pull/76#issuecomment-28942928) to @rickynils.

I'm happy to contribute the change if someone suggests to me which of these approaches seems like the better option.

ceedubs:
Resolves #401.

### DIFF
--- a/src/main/scala/org/scalacheck/Arbitrary.scala
+++ b/src/main/scala/org/scalacheck/Arbitrary.scala
@@ -68,7 +68,7 @@ object Arbitrary extends ArbitraryLowPriority with ArbitraryArities {
 
 /** separate trait to have same priority as ArbitraryArities */
 private[scalacheck] sealed trait ArbitraryLowPriority {
-  import Gen.{const, choose, sized, frequency, oneOf, buildableOf, resize}
+  import Gen.{const, choose, sized, frequency, oneOf, buildableOf}
 
   /** Creates an Arbitrary instance */
   def apply[T](g: => Gen[T]): Arbitrary[T] = new Arbitrary[T] {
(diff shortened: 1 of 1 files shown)


exoego:
@ceedubs 
I would like to have this PR merged.

Just 1 build terminated due to `The command "curl https://raw.githubusercontent.com/scala-native/scala-native/master/scripts/travis_setup.sh | bash -x" failed and exited with 100 during .`

Perhaps restarting Travis CI is just fine.
We can restart CI by closing this PR and reopening.

ashawley:
There's also should be a way in Travis to hit a restart button on individual failed builds, too.

ceedubs:
The 2.13 build failed with no output: https://travis-ci.org/rickynils/scalacheck/jobs/437892113

I closed and reopened once. Maybe someone with permissions could restart just that sub-build?

dwijnand:
This looks simple to review and merge.  It's a shame it's been left hanging.

PR #409, opened 2018-05-12
Add Gen.either

ceedubs:
And make the Arbitrary instance for Either delegate to it.


non:
If we merge #494 we can close this.

non:
Thanks @ceedubs! Sorry for how long this sat around.

Issue #410, opened 2018-05-15
publish for Scala 2.13.0-M4?


SethTisue:
M4 JARs are on Maven Central now

in the 2.13 community build we are using https://github.com/SethTisue/scalacheck/tree/newCollections (forked from @lrytz who did most of the work)

rickynils:
#411 changes the API, due to the type changes (`Stream -> LazyList`, `Traversable -> Iterable`). I'm not sure how we should handle the releases and versioning from here on. It feels wrong to just release #411 as another `1.14.0` build. @SethTisue @lrytz Any thoughts on that?

lrytz:
I think the we should probably revert the `Stream -> LazyList` change, but instaed add built-in generators for `LazyList` in the 2.13 version.

For the `Traversable -> Iterable` change, we can basically introduce a new type alias which, in 2.10/2.11/2.12 would be `Traversable`, and 2.13 would be `Iterable`.

Both of the above can be done with version-dependent source directories. This is a bit annoying to have, but I think it's better than having a separate branch for 2.13.

Let me know what you think, I'll put it in the PR.

lrytz:
Once released, re-enable MiMa for 2.13.0-M4 in build.sbt (mimaPreviousArtifacts).

ashawley:
> I think the we should "revert" the `Stream` → `LazyList` change, and instead add built-in generators for `LazyList` in the 2.13 version.

Changing those as described would would be necessary to publish 1.14.0 for 2.13.0-M4.  Otherwise,  it doesn't seem like the 1.14.0 release would be source compatible.

lrytz:
That was done in the PR (https://github.com/rickynils/scalacheck/pull/411) (which is still ready for review :relaxed:).

ashawley:
Ah, ok. Sorry, it wasn't last I checked. I'll look again.

ashawley:
The signature for `Shrink` was still using `LazyList` instead of `Stream` in the 2.13 build.

ashawley:
Since Mima can't really check compatability with new versions of Scala, one other way to verify compatability is to try to export Scalacheck's test suite for 1.14.0 and try to compile it with this new 1.14.0 build (alongside the existing binaries).  This assumes Scalacheck has sufficient test coverage of itself, but that may be a flawed assumption.  

[...]

PR #411, opened 2018-05-15
Changes to compile with Scala 2.13.0-M4

lrytz:


lrytz:
Yay, this is green now! @ashawley raises concerns about binary compatibility over at https://github.com/scala/scala-xml/pull/222. Have to check in detail and see if we can keep doing 1.14.x, or need to move to 1.15.

lrytz:
I went through all the changes again with focus on binary and source compatibility. It looks to me like we can continue release 0.14.x releases for 2.10-2.12 after merging this PR, they should be backwards compatible.

I think this PR is ready now. Review by @rickynils :-)

.travis.yml
```diff
@@ -46,3 +49,5 @@ matrix:
   exclude:
     - scala: 2.10.7
       env: PLATFORM=js  SBT_PARALLEL=true  WORKERS=1 DEPLOY=true SCALAJS_VERSION=1.0.0-M3
+    - scala: 2.13.0-M4
+      env: PLATFORM=js  SBT_PARALLEL=true  WORKERS=1 DEPLOY=true SCALAJS_VERSION=1.0.0-M3
```
raboof: This exclude doesn't make sense like this anymore since the change from scala 2.13.0-M3 to 2.13.0-M4, right?

build.sbt
```diff
@@ -78,7 +91,11 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
   // don't use fatal warnings in tests
   scalacOptions in Test ~= (_ filterNot (_ == "-Xfatal-warnings")),
 
-  mimaPreviousArtifacts := Set("org.scalacheck" %% "scalacheck" % "1.14.0"),
+  mimaPreviousArtifacts := {
+    // TODO: re-enable MiMa for 2.13.0-M4 once there is a release out
```
raboof: Maybe add a ticket to keep track of this?

src/main/scala/org/scalacheck/Prop.scala
```diff
@@ -765,8 +766,8 @@ object Prop {
     /*
      * Returns the first failed result in Left or success in Right.
      */
-    def getFirstFailure(xs: Stream[T]): Either[(T,Result),(T,Result)] = {
-      assert(!xs.isEmpty, "Stream cannot be empty")
+    def getFirstFailure(xs: LazyList[T]): Either[(T,Result),(T,Result)] = {
```
raboof: Is this binary compatible?

.travis.yml
lrytz: Not sure I understand.. This exclude skips building with Scala.js 1.0.0-M3 on Scala 2.13.0-M4, because there's no such Scala.js release yet.

src/main/scala/org/scalacheck/Prop.scala
lrytz: Yes, because `LazyList` is an alias for `Stream` when building on 2.10-2.12 (https://github.com/rickynils/scalacheck/blob/543b4056bffdb684faec49e8afb09aee4c24eb2f/src/main/scala-2.13-/org/scalacheck/ScalaVersionSpecific.scala#L19-L20)

build.sbt
lrytz: Added a comment https://github.com/rickynils/scalacheck/issues/410#issuecomment-391971774

.travis.yml
raboof: A got it :+1:

src/main/scala/org/scalacheck/Prop.scala
raboof: :+1:

src/main/scala/org/scalacheck/Shrink.scala
```diff
@@ -14,59 +14,60 @@ import language.higherKinds
 import util.Buildable
 import util.SerializableCanBuildFroms._
 import scala.concurrent.duration.{Duration, FiniteDuration}
+import ScalaVersionSpecific._
 
 sealed abstract class Shrink[T] extends Serializable {
-  def shrink(x: T): Stream[T]
+  def shrink(x: T): LazyList[T]
```
ashawley: So the signature for `Shrink` changes here, so existing 1.14.0 tests (or libraries) that defined shrinkers would no longer compile.

lrytz: The signature is actually the same on 2.10-2.12, because `LazyList` is an alias for `Stream` there: https://github.com/rickynils/scalacheck/blob/543b4056bffdb684faec49e8afb09aee4c24eb2f/src/main/scala-2.13-/org/scalacheck/ScalaVersionSpecific.scala#L19-L20

ashawley: Here's an example shrinker:

```scala
object CustomShrink {
  implicit val dumbIntShrinker: Shrink[Int] = Shrink {
    case _ => Stream.empty
  }
}
```

Publishing this branch as 1.14.0-543b405, the example compiles on 2.10, 2.11, and 2.12, but for 2.13.0-M4:

```
[error] src/main/scala/CustomShrink.scala:5:22: polymorphic expression cannot be instantiated to expected type;
[error]  found   : [A]scala.collection.immutable.Stream[A]
[error]  required: LazyList[Int]
[error]     case _ => Stream.empty
```

lrytz: Thank you for pointing this out.

Do poeple actually write shrinkers "ahead of time", at the same time when writing `Arbitrary` instances, and put them into the code base?

We could make the `LazyList` aliases public on 2.10-2.12, then poeople could use these. But I think this gets even more confusing. Already having the `LazyList` alias in the signatures on 2.10-2.12 is a bit problematic.

We could keep using `Stream`, but that would lead to deprecation warnings on 2.13.

ashawley: Yeah, writing one's own `Shrink` is documented.

ashawley: Aren't we already emitting warnings with `Traversable`?

lrytz: There are a number of uses of `Traversable` in the API, but I think they should not lead to deprecation warnings for the users. In 2.13, `Traversable` is a (deprecated) alias to `Iterable`.

So let's say scalacheck defines a method

```
scala> def foo: Traversable[Int] = List(1,2,3)
                ^
       warning: type Traversable in package scala is deprecated (since 2.13.0): Use Iterable instead of Traversable
foo: Traversable[Int]
```

When compiling scalacheck on 2.13, this will give a deprecation warning. But for users it won't:

```
scala> foo
res0: Traversable[Int] = List(1, 2, 3)

scala> foo.map(_ + 1)
res1: Iterable[Int] = List(2, 3, 4)

scala> foo: Iterable[Int]
res2: Iterable[Int] = List(1, 2, 3)
```

ashawley: Ah, ok.  I see the distinction.  Thanks for explaining.

For now, preserve `Stream` in `Shrink`, and then migrate to `LazyList` in 1.15.0?

lrytz:
Pushed a change by @ashawley to go back to using `Stream` when defining `Shrink`ers. We can move to `LazyList` in a scalacheck 1.15

ashawley:
I'd say this is good to publish for 2.13.0-M4.  

Since there are long-term implications with 2.13 that require further consideration going forward (like, "What should go in 1.15.0?"), maybe hold off on the merge?

rickynils:
@ashawley I'v rebased and merged this to the 1.14.0_sonatype branch and published the build. I'm not sure I see why we should hold off the merge to master, though?

ashawley:
This PR currently represents the conservative set of changes for making a 1.14.0 release for 2.13.  Publishing was important for getting a ScalaCheck release out for the 2.13.0-M4 milestone .   That's all that's been really reviewed, though.  

What wasn't probably reviewed, was the first half of these commits of this PR containing the original re-implementation of ScalaCheck's use of collections for 2.13.  Those modifications have compatibility breaking changes, but they would seem like the better base with which to move forward on.

lrytz:
In the current form this PR doesn't have breaking changes, those parts of the earlier commits are reverted in later commits. I can squash things if that's desired.

Changes such as moving to use `LazyList` instead of `Stream` in shrinkers on 2.13 can be done in a 1.15 release.

build.sbt
```diff
@@ -54,18 +67,18 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
...
-    scalaBinaryVersion.value match {
-      case "2.10" => Seq("-Xlint")
-      case "2.11" => Seq("-Xlint", "-Ywarn-infer-any", "-Ywarn-unused-import")
-      case _      => Seq("-Xlint:-unused", "-Ywarn-infer-any", "-Ywarn-unused-import", "-Ywarn-unused:-patvars,-implicits,-locals,-privates,-explicits")
+    val modern = Seq("-Xlint:-unused", "-Ywarn-infer-any", "-Ywarn-unused-import", "-Ywarn-unused:-patvars,-implicits,-locals,-privates,-explicits")
+    scalaMajorVersion.value match {
+      case 10 => Seq("-Xfatal-warnings", "-Xlint")
+      case 11 => Seq("-Xfatal-warnings", "-Xlint", "-Ywarn-infer-any", "-Ywarn-unused-import")
+      case 12 => "-Xfatal-warnings" +: modern
+      case 13 => modern
```
jozic: does it mean 2.13 enables `-Xfatal-warnings` by default?

lrytz: No :-) this disables fatal warnings on 2.13 because there are some deprecation warnings due to cross-building (usage of Stream)

jozic: got it, one way could be to add silencer for that particular

mcanlas:
Are there any blockers for this PR?

Scala 2.13.0-M5 is out now.

ashawley:
M5 is out, but I'm not sure it's minted, yet.

For the record, this was merged to the 1.14.0_sonatype branch in #410, but just not master, yet.  Since development isn't active, this is effectively a longer-running branch/PR for 2.13.

lrytz:
It would be great to move 2.13 cross-building forward. As mentioned in a [recent announcement](https://contributors.scala-lang.org/t/2-13-0-rc1-timing-january-2019/2375), one goal of the postponed RC1 is to make sure we have time to get the OSS ecosystem up and running. Scalacheck is obviously very important as it's one of the roots.

I'll update this PR to M5.

ashawley:
Seth got ScalaCheck most of the way there for 2.13.0-M5, see #427.

rickynils:
I have merged `1.14.0_sonatype` to `master`, so this PR is now effectively merged to `master` too. Closing this.

Issue #355, opened 2017-08-26
Pick does not choose randomly when choosing all elements


ssanj:
Pick is defined as:

```
def pick[T](n: Int, l: Iterable[T]): Gen[Seq[T]]
```
I use ```pick```  to  randomly choose 3 values from a range of 1 to 5 with the following:

```
scala> List.fill(10)(Gen.pick(3, (1 to 5)).sample.get).mkString("\n")
res142: String =
ArrayBuffer(1, 2, 5)
ArrayBuffer(1, 5, 3)
ArrayBuffer(4, 2, 5)
ArrayBuffer(1, 5, 4)
ArrayBuffer(1, 2, 5)
ArrayBuffer(5, 2, 4)
ArrayBuffer(4, 2, 5)
ArrayBuffer(1, 4, 5)
ArrayBuffer(5, 2, 3)
ArrayBuffer(1, 5, 4)
```

We see that the 3 values are chosen randomly.

If I ask for all 5 values to be chosen randomly:

```
scala> List.fill(10)(Gen.pick(5, (1 to 5)).sample.get).mkString("\n")
res143: String =
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
ArrayBuffer(1, 2, 3, 4, 5)
```

I keep getting the same values.

Is this the expected behaviour? It seems like unless the ```n``` supplied to pick is less than the size of the iterable ```i```, it does not return random values.

ashawley:
It isn't released, yet, but it was fixed in #333.

ssanj:
I built this locally and it doesn't seem to have solved the issue. The problem is in the ```if``` condition, where it returns the exact content of the iterable __l__ when ```count <= n```  where, __count__ is the number of elements in __l__ and __n__ is the number of elements to pick:

```
      while (it.hasNext) {
        val t = it.next
        count += 1
        if (count <= n) {
          buf += t //adds all elements of l to buf while count <= n; which is what we requested - all elements)
        } else {
          val (x, s) = seed.long
[...]

PR #417, opened 2018-08-10
Improve doc for Gen.pick

ashawley:
This is in response to concerns raised about `Gen.pick` in #355.

`Gen.pick`, `Gen.someOf`, and `Gen.atLeastOne` will give a randomly chosen subset of elements, but not in a random order.  This is because `Gen.pick` is implemented with reservoir sampling but without a shuffle operation.  Adding shuffling might be possible, but it could impact performance and would violate the "do one thing well" principle of `Gen.pick`.

For now, just document the situation that results aren't permuted in the User Guide and in the API docs.

### DIFF
--- a/doc/UserGuide.md
+++ b/doc/UserGuide.md
@@ -484,6 +484,39 @@ examples.
 There is also `Gen.nonEmptyContainerOf` for generating non-empty containers, and
 `Gen.containerOfN` for generating containers of a given size.
 
+To generate a container by picking an arbitrary number of elements use
+`Gen.someOf`, or by picking one or more elements with
+`Gen.atLeastOne`.
+
+```scala
+val zeroOrMoreDigits = Gen.someOf(1 to 9)
+
+val oneOrMoreDigits = Gen.atLeastOne(1 to 9)
+```
+
+Here are generators that randomly pick `n` elements from a container
+with `Gen.pick`:
+
+```scala
+val fiveDice: Gen[Seq[Int]] = Gen.pick(5, 1 to 6)
+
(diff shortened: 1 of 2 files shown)


doc/UserGuide.md
```diff
@@ -486,6 +486,40 @@ examples.
...
+
+```scala
+val fiveDice: Gen[Seq[Int]] = Gen.pick(5, 1 to 6)
+
+val threeLetters: Gen[Seq[Char]] = Gen.pick(3, 'A' to 'Z')
+```
+
+Note that `Gen.someOf`, `Gen.atLeastOne`, and `Gen.pick` only randomly
+select elements, they do not generate permutations of the selected
+elements in different orders.  Selection is random, but the order is
```
pford19: Observed behavior of `pick` is that `pick(2, 1 to 10)` generates (for large samples) 73 distinct 2-lists, out of a total of 90 possible. 73 < 90 is consistent with the remark that `pick` does not, in general, permute all its results. 

But since there are only 45 (=10*9/2) distinct 2-sets from (1 to 10), 73 > 45 implies that `pick` __does permute__ some of its results, so the order is __not deterministic__. For instance `List(4, 5)` and `List(5, 4)` are both generated by `pick(2, 1 to 10)`.

ashawley: True, using the word "deterministic" is confusing.   I'll remove it.

ashawley:
This issue came up again in the Gitter channel, yesterday.  

The Scaladoc changes won't help right away, not until a new release is published, but the User Guide would be updated right away.  The latter probably has a better chance of getting noticed.

PR #420, opened 2018-09-10
Fix the scaladocs of Gen.mapOf

hosamaly:
The methods `containerOf`, `containerOfN`, and `nonEmptyContainerOf` all accept two type parameters, not three.

### DIFF
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -66,7 +66,7 @@ sealed abstract class Prop extends Serializable { self =>
     this
   )
 
-  /** Convenience method that checks this property with the given parameters
+  /** Convenience method that checks this property
    *  and reports the result on the console. Should only be used when running
    *  tests interactively within the Scala REPL.
    *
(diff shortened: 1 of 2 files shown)


src/main/scala/org/scalacheck/Prop.scala
```diff
@@ -66,7 +66,7 @@ sealed abstract class Prop extends Serializable { self =>
     this
   )
 
-  /** Convenience method that checks this property with the given parameters
+  /** Convenience method that checks this property
```
ashawley: Maybe it should read "check this property with the [default] parameters"?

hosamaly: Maybe. I noticed the link to the `[default] parameters` on line 73, so I thought the original author preferred to highlight it separately. Thus, I made it consistent with the docs of the next method (on line 77).

I'm happy to change it if you wish, but then I think I should change lines 69, 73, and 77.

ashawley: Probably fine to mention it uses the default parameters _and_ links to the default parameters later.

PR #423, opened 2018-09-14
Changes to scalacheck to cross compile with Dotty

allanrenucci:
I left out https://github.com/dotty-staging/scalacheck/commit/00f01afeb2e76f0d2a4266fe60900699a87453c2 on purpose because it is a binary incompatible change. Apart from that `scalacheck` cross compiles with Dotty


jvm/src/test/scala/org/scalacheck/GenSpecification.scala
```diff
@@ -400,7 +400,7 @@ object GenSpecification extends Properties("Gen") {
   )
 
   property("22 field case class works") = forAll(Gen.resultOf(Full22.tupled)){
-    Full22.unapply(_).get.isInstanceOf[Tuple22[_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]]
+    _.isInstanceOf[Full22]
```
ashawley: This won't compile if ScalaCheck's tuple support is defective, so could probably just be written simply like the rest of the similar tests found earlier in the suite (`resultOf1`, `resultOf2`, ...):

```scala
  property("22 field case class works") =
    forAll(Gen.resultOf(Full22.tupled)) { _ => true }
```

src/main/scala/org/scalacheck/rng/Seed.scala
```diff
@@ -98,10 +98,12 @@ sealed abstract class Seed extends Serializable {
 
 object Seed {
 
-  private case class apply(a: Long, b: Long, c: Long, d: Long) extends Seed {
+  private case class SeedImpl(a: Long, b: Long, c: Long, d: Long) extends Seed {
     override def toString: String = s"""Seed.fromBase64("$toBase64")"""
   }
 
+  def apply(a: Long, b: Long, c: Long, d: Long): Seed = SeedImpl(a, b, c, d)
+
```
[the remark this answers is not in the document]
allanrenucci: Ideally it should probably be:
```scala
final class Seed private (a: Long, b: Long, c: Long, d: Long) extends Serializable {
  ...
}

object Seed {
  private def apply(a: Long, b: Long, c: Long, d: Long) = new Seed(a, b, c, d)
  ...
}
```
But I couldn't make the serialization tests pass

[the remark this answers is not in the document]
ashawley: Yes, good point.  Looks like `SeedImpl` is the more common idiom.

er1c:
@rickynils this would be great! *bump*!

smarter:
All comments against this PR were addressed 9 months ago, and yet it wasn't merged or closed, what's the hold up ? I've now rebased this PR to fix conflicts, while I was at it I removed one change which is no longer needed (Seed#apply) and added two extra commits to let Dotty compile the code without `-language:Scala2`. Can I expect it to be merged this time around ?

ashawley:
The `Seed.apply` change would have technically broken binary compatibility.  I can't speak for anyone else, but I didn't finish reviewing the changes for that reason.  Since it's now no longer needed, I suppose there's a chance that these fixes for Dotty could be included in a 1.14.1 bug fix release, IMO.  With that in mind, I'll try to study the changes again and see if that's the case.

ashawley:
Seems like all of this could be included a 1.14.1 bug fix release to me.  I think the change to `CmdLineParser` is an internal change, so it shouldn't be of any consequence to the user.

src/main/scala/org/scalacheck/Test.scala
```diff
@@ -225,7 +225,7 @@ object Test {
     ) throw new IllegalArgumentException("Invalid test parameters")
   }
 
-  private[scalacheck] lazy val cmdLineParser = new CmdLineParser {
+  private[scalacheck] object cmdLineParser extends CmdLineParser {
```
ashawley: Should probably be renamed `CmdLineParser` per naming conventions elsewhere for class level objects in the code base, even if _uniform access_ would suggest otherwise.  This requires renaming a bunch of references elsewhere, but that probably helps explain why the`reflectiveCalls` import is removed in those files.

smarter: Done!

smarter:
Are source-incompatible but binary-compatible changes allowed in the 1.14 branch ? One way to obtain the same result as https://github.com/dotty-staging/scalacheck/commit/00f01afeb2e76f0d2a4266fe60900699a87453c2 would be to simply remove the `implicit` from `implicit def BooleanOperators(b: => Boolean): ExtendedBoolean`, but existing code might have to replace `import Prop.BooleanOperators` by `import Prop.propBoolean`. If that's not OK, then is there a plan to create a branch for the development of 1.15 ?

ashawley:
As described, that sounds like a change like that would be pretty disruptive to users in a 1.14.1 release.

I wonder if you should create an issue (or issues) to discuss the various Dotty changes you need in ScalaCheck.  One assumes there will need to be a lot of coordination and investigation.

smarter:
Besides this PR, the BooleanOperators thing is the only necessary change I'm aware of, we switched the Functional Programming course at EPFL to Dotty last year, the course introduces students to property checking with scalacheck (like the mooc), and those changes were the only one we needed to get everything working.

smarter:
Anything else needed to get this merged ?

smarter:
> One way to obtain the same result as dotty-staging@00f01af would be to simply remove the implicit from `implicit def BooleanOperators(b: => Boolean): ExtendedBoolean`

As a first step towards this, I've opened a PR to deprecate `BooleanOperators`: https://github.com/typelevel/scalacheck/pull/498

ashawley:
Unfortunately, we merged #498 first so there's a conflict.  Fortunately, the conflict is just a single line.  Would you be able to rebase?  I also wonder if the new version of MiMa raises anything else.

smarter:
Rebased.

ashawley:
Strange errors from MiMa:

```
[error]  * method <clinit>()Unit in class java.lang.Object is inaccessible in current version, it must be public.
[error]    filter with: ProblemFilters.exclude[InaccessibleMethodProblem]("java.lang.Object.<clinit>")
[error]  * method <clinit>()Unit in class java.lang.Object is inaccessible in current version, it must be public.
[error]    filter with: ProblemFilters.exclude[InaccessibleMethodProblem]("java.lang.Object.<clinit>")
```

It happens for Scala 2.11 and 2.12 for both Java and Scala.js 6.0.  I'm not sure what to make of this.

SethTisue:
@dwijnand is that something you've seen before?

smarter:
Try restarting the build, travis flakiness manifests in weird ways sometimes

ashawley:
Unfortunately, this also happens for me locally, as well.

dwijnand:
No, that's a new one.

ashawley:
I was able to bisect it to the "Stop using reflective calls" commit.  There's a lot of changes in that commit, but the change of`Test.cmdLineParser` from a lazy val to an object is probably the issue.

SethTisue:
rebase, now that #521 is merged?

ashawley:
Would @non weigh-in on the binary-compat change?  I've been meaning to explore it, but haven't because I predict he may have draw a harder line on this then I had done so previously.

PR #430, opened 2018-09-21
work around scala/bug#11125

SethTisue:
this is for the Scala community build, where we want to be able
to build everything on JDK 11 to help test Scala

references scala/bug#11125


SethTisue:
the CI failure looks like it just needs a committer to hit "rebuild" on the single job that failed

PR #433, opened 2018-10-07
use valueOf for construction class Character

magnolia-k:
Since Java 9, constructor Character in class Character is deprecated.


ashawley:
Fortunately, `Character.valueOf` has existed since Java 1.5:

https://docs.oracle.com/javase/9/docs/api/java/lang/Character.html#valueOf-char-

ashawley:
Maybe we should update the Travis build to compile on openjdk11 for 2.12 and later?

PR #434, opened 2018-10-07
Add openjdk11 to Travis build

ashawley:
Java 11 is out with long-term support.  Adding it to the build adds the build count in Travis from 21 to 31, and the running clock time goes from 30 min to 40 min.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ scala:
   - 2.13.0-M3
 jdk:
   - oraclejdk8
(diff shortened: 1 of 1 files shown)


ashawley:
It works, so that's good news.  It does reproduce the deprecation warning in #433.

However, it does grow the build (see above).  Maybe drop support for 2.10 and 2.11 in a future version of ScalaCheck, and then add openjdk11?  Unless, there are major issues found with Java 11.

PR #436, opened 2018-10-14
Update dead link to grepcode

aki-ks:
Grepcode is dead. openjdk.java.net should be more lasting.

### DIFF
--- a/src/main/scala/org/scalacheck/Arbitrary.scala
+++ b/src/main/scala/org/scalacheck/Arbitrary.scala
@@ -121,7 +121,7 @@ private[scalacheck] sealed trait ArbitraryLowPriority {
 
   /** Arbitrary instance of Char */
   implicit lazy val arbChar: Arbitrary[Char] = Arbitrary {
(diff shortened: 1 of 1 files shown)


aki-ks:
That's strange. I've just changed a comment. How can this cause the build to fail?

Hydrotoast:
Hi @aki-ks , I checked the build failure. I ran into the same issue in #438, which is a flaky test that occurs with a frequency of 1 / 1000 per build job (I believe there are ~10-15 build jobs per PR). The PR includes a fix in the second commit.

Due to the rarity of the flakiness, you can trigger a new build that should pass with a high probability of success. A build should be triggered when this PR detects a change in the source branch of the merge, which appears to be `patch-1` of your repository. 

I believe you can amend the the timestamp of the commit without changing the diff or the message and force push to your `patch-1` branch.  The following should work, but be sure that you are on the branch that has the desired change. :)
```
git commit --amend --no-edit
git push -f origin patch-1
```

Issue #317, opened 2017-03-05
Tests generating invalid inputs (value 0)


nlw0:
I am working on a software for logic inference, where I am processing lists of non-zero values. My test looks like:

```
val big_literal = Gen.oneOf(Gen.choose(-50, -1), Gen.choose(1, 50))
val big_clause = Gen.listOf(big_literal).map(_.distinct)
val big_clause_list = Gen.listOf(big_clause)

forAll(big_clause_list) { lol =>
      println(s"lol: $lol")
...
}
```

This always breaks for me, with Scalacheck generating shrinked lists that contain the value 0, which is invalid. If I switch to `forAllNoShrink` all goes well.

I was very lost on that for a while, it was quite surprising to see the value '0' in my tests while I was very careful to generate the data without it, and even tried to implement that in different ways. I think shrinking is a great feature, but I got bitten hard this time... Shouldn't there be some more care to prevent something like that?

rickynils:
This is a known limitation of ScalaCheck, shrinking will not always respect the generator bounds.

You can add a pre-condition to you property with ScalaCheck's implication operator (`==>`) to avoid this if you like. That would also make it clear that your implementation has a pre-condition that needs to be satisfied before using it.

nlw0:
Thanks. I ended up writing a Shrink and "specializing" my Ints using a case class (because I don't know how to use stuff like refined types), but I am definitely interested in learning how to use the library better and try something like that. I don't mean to be a whiny user, and shrinking actually helped me to find a bug in my code later. I understand it is difficult to implement this, and I've been thinking about the whole problem of generating the data while supporting the constraints. I noticed the whole process, both generating and filtering, could be implemented as a kind of Markov-chain Monte Carlo, where samples are generated by transforming other valid samples instead of always filtering something much more general. You would have a "Mutate" like a "Shrink", and go on applying transforms that would be more likely within the constraints, until you stop and produce a new sample. That might make sampling more efficient in some cases. Have you ever considered something like that?

Kevin-Lee:
@rickynils Could you please tell me if this will be fixed to always respect the generator bounds or it is an intended behavior so will remain as is?
> This is a known limitation of ScalaCheck, shrinking will not always respect the generator bounds.

rickynils:
@Kevin-Lee I consider it something that should eventually be fixed, but it might require larger changes in ScalaCheck, and no one is currently working on this.

charleso:
[...]

Issue #260, opened 2016-09-06
Gen.choose not working as expected


alxest:
In below code, I expect x and y to pass assertion, but it fails. (x and y chosen from [0,8])
Is this intended behavior?
```
import org.scalacheck.Properties
import org.scalacheck.Prop.forAll
import org.scalacheck._

object PascalSpec extends Properties("pascal") {
  def pascal1(x: Int, y: Int): Int = {
    if(x == 0 || y == 0 || y == x) 1
    else pascal1(x-1, y) + pascal1(x-1, y-1)
  }

  def pascal2(x: Int, y: Int): Int = {
    def factorial(n: Int): Int =
      if(n == 0 || n == 1) 1 else factorial(n-1)
    factorial(x) / factorial(y) / factorial(x-y)
  }

  val smallIntPair = for {
    x <- Gen.choose(5, 8)
    y <- Gen.choose(5, 8)
  } yield (x, y)

  property("spec") = forAll(smallIntPair) {
    (n: (Int, Int)) => {
      val (x, y) = n
      println("{" + x + " " + y + "}")
      assert(5 <= x && x <= 8)
      assert(5 <= y && y <= 8)
      val ret = (pascal1(x,y) == pascal2(x,y))
      ret
    }
  }
}
```
Replacing       
```
val ret = (pascal1(x,y) == pascal2(x,y))
      ret
```
into 
```
true
```
makes assertion pass.
I am using scala 2.12.8 and I uploaded the code [here](https://github.com/alxest/scalacheck_question/blob/master/src/test/scala/test.scala). It should be reproducible.

non:
Can you paste the output from the failing test/assertion?

rickynils:
Your implementation (or property) seems to be wrong. See:

```
scala> def pascal1(x: Int, y: Int): Int = {
     |     if(x == 0 || y == 0 || y == x) 1
     |     else pascal1(x-1, y) + pascal1(x-1, y-1)
     |   }
pascal1: (x: Int, y: Int)Int

scala> def pascal2(x: Int, y: Int): Int = {
     |     def factorial(n: Int): Int =
     |       if(n == 0 || n == 1) 1 else factorial(n-1)
     |     factorial(x) / factorial(y) / factorial(x-y)
     |   }
pascal2: (x: Int, y: Int)Int

scala> pascal1(8,7)
res1: Int = 8

scala> pascal2(8,7)
res2: Int = 1
```

[...]

Issue #129, opened 2015-01-07
Gen.suchThat not respected by shrinking after Gen.map


2rs2ts:
My team has a library which contains a bunch of predefined generators. Here's [a simple one](https://github.com/paypal/cascade/blob/develop/common/src/test/scala/com/paypal/cascade/common/tests/scalacheck/scalacheck.scala#L55-L56):

```scala
lazy val genNonEmptyAlphaStr: Gen[String] = Gen.nonEmptyListOf(Gen.alphaChar).map(_.mkString)
```

Unfortunately when I go to use it in practice, this doesn't get respected at all. (We use specs2 along with ScalaCheck, so bear with me.)

```scala
import org.scalacheck.Prop._
import org.specs2.{ScalaCheck, SpecificationLike}
import com.paypal.cascade.common.tests.scalacheck._

class ShrinkTestSpecs
  extends SpecificationLike
  with ScalaCheck { def is = s2"""
  lalal ${test}
  """

  def fail = throw new RuntimeException("aaa!")

  def test = {
    forAll(genNonEmptyAlphaStr) { s =>
      fail
      true must beTrue
    }
  }
}
```

This results in some output like

```
Testing started at 1:37 PM ...

A counter-example is '': java.lang.RuntimeException: aaa! (after 0 try)
java.lang.RuntimeException: aaa!
```

If I use `forAllNoShrink`, or if I write `genNonEmptyAlphaStr.suchThat(_.size > 0)`, then there are no problems and I get single-letter strings. If I use `Gen.nonEmptyListOf(Gen.alphaChar)` (in other words, I don't map on it,) then I also have no problems. It's only after I `.map(_.mkString)` that it will shrink to the empty string.

tl;dr `Gen.map` invalidates `suchThat`.

rickynils:
This is a known issue. With the current design it is not possible retain the `suchThat` filter after a `map`. This will hopefully change in ScalaCheck 2, where the generators have been redesigned to better handle these cases.

mlangc:
Hmm, I guess this is the same issue then:

Using *scalacheck-1.12.2* with *scalatest-2.2.5* and *scala-2.11.6* executing
```scala
 "Test scalacheck" in {
    val gen = for {
      n <- Gen.choose(1, 3)
      list <- Gen.listOfN(n, Gen.choose(0, 9))
    } yield list
    
    forAll(gen) { list =>
      list should not be (empty)
[...]

PR #440, opened 2018-11-23
Remove shrinking from gen-based Prop.forAll

charleso:
Shrinking is only safe when used with Arbitrary. Generators can have a smaller range than the global shrinking, which results in invalid shrunken values.

This fixes issues like #129, #260, #317, #370, #442 and #443.

I suspect this is going to be controversial and I completely understand if it's not accepted. I'm opening this to show how I suggest it should be tackled, but it obviously affects people who are currently relying on this behaviour.

EDIT: Note that Haskell's QuickCheck `forAll` [doesn't shrink in this instance](http://hackage.haskell.org/package/QuickCheck-2.12.4/docs/src/Test.QuickCheck.Property.html#forAll) either

### DIFF
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -804,106 +804,97 @@ object Prop {
 
   }
 
-  /** Universal quantifier for an explicit generator. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for an explicit generator. */
   def forAll[T1,P](
     g1: Gen[T1])(
     f: T1 => P)(implicit
     p: P => Prop,
-    s1: Shrink[T1],
     pp1: T1 => Pretty
-  ): Prop = forAllShrink[T1,P](g1, shrink[T1])(f)
+  ): Prop = forAllNoShrink(g1)(f)
 
-  /** Universal quantifier for two explicit generators. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for two explicit generators. */
   def forAll[T1,T2,P](
     g1: Gen[T1], g2: Gen[T2])(
     f: (T1,T2) => P)(implicit
     p: P => Prop,
-    s1: Shrink[T1], pp1: T1 => Pretty,
-    s2: Shrink[T2], pp2: T2 => Pretty
-  ): Prop = forAll(g1)(t => forAll(g2)(f(t, _:T2)))
+    pp1: T1 => Pretty,
+    pp2: T2 => Pretty
+  ): Prop = forAllNoShrink(g1, g2)(f)
 
-  /** Universal quantifier for three explicit generators. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for three explicit generators. */
   def forAll[T1,T2,T3,P](
     g1: Gen[T1], g2: Gen[T2], g3: Gen[T3])(
     f: (T1,T2,T3) => P)(implicit
     p: P => Prop,
-    s1: Shrink[T1], pp1: T1 => Pretty,
-    s2: Shrink[T2], pp2: T2 => Pretty,
-    s3: Shrink[T3], pp3: T3 => Pretty
-  ): Prop = forAll(g1)(t => forAll(g2,g3)(f(t, _:T2, _:T3)))
+    pp1: T1 => Pretty,
+    pp2: T2 => Pretty,
+    pp3: T3 => Pretty
+  ): Prop = forAllNoShrink(g1, g2, g3)(f)
 
-  /** Universal quantifier for four explicit generators. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for four explicit generators. */
   def forAll[T1,T2,T3,T4,P](
     g1: Gen[T1], g2: Gen[T2], g3: Gen[T3], g4: Gen[T4])(
     f: (T1,T2,T3,T4) => P)(implicit
     p: P => Prop,
-    s1: Shrink[T1], pp1: T1 => Pretty,
-    s2: Shrink[T2], pp2: T2 => Pretty,
-    s3: Shrink[T3], pp3: T3 => Pretty,
-    s4: Shrink[T4], pp4: T4 => Pretty
-  ): Prop = forAll(g1)(t => forAll(g2,g3,g4)(f(t, _:T2, _:T3, _:T4)))
+    pp1: T1 => Pretty,
+    pp2: T2 => Pretty,
+    pp3: T3 => Pretty,
+    pp4: T4 => Pretty
+  ): Prop = forAllNoShrink(g1, g2, g3, g4)(f)
 
-  /** Universal quantifier for five explicit generators. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for five explicit generators. */
   def forAll[T1,T2,T3,T4,T5,P](
     g1: Gen[T1], g2: Gen[T2], g3: Gen[T3], g4: Gen[T4], g5: Gen[T5])(
     f: (T1,T2,T3,T4,T5) => P)(implicit
     p: P => Prop,
-    s1: Shrink[T1], pp1: T1 => Pretty,
-    s2: Shrink[T2], pp2: T2 => Pretty,
-    s3: Shrink[T3], pp3: T3 => Pretty,
-    s4: Shrink[T4], pp4: T4 => Pretty,
-    s5: Shrink[T5], pp5: T5 => Pretty
-  ): Prop = forAll(g1)(t => forAll(g2,g3,g4,g5)(f(t, _:T2, _:T3, _:T4, _:T5)))
+    pp1: T1 => Pretty,
+    pp2: T2 => Pretty,
+    pp3: T3 => Pretty,
+    pp4: T4 => Pretty,
+    pp5: T5 => Pretty
+  ): Prop = forAllNoShrink(g1, g2, g3, g4, g5)(f)
 
-  /** Universal quantifier for six explicit generators. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for six explicit generators. */
   def forAll[T1,T2,T3,T4,T5,T6,P](
     g1: Gen[T1], g2: Gen[T2], g3: Gen[T3], g4: Gen[T4], g5: Gen[T5], g6: Gen[T6])(
     f: (T1,T2,T3,T4,T5,T6) => P)(implicit
     p: P => Prop,
-    s1: Shrink[T1], pp1: T1 => Pretty,
-    s2: Shrink[T2], pp2: T2 => Pretty,
-    s3: Shrink[T3], pp3: T3 => Pretty,
-    s4: Shrink[T4], pp4: T4 => Pretty,
-    s5: Shrink[T5], pp5: T5 => Pretty,
-    s6: Shrink[T6], pp6: T6 => Pretty
-  ): Prop = forAll(g1)(t => forAll(g2,g3,g4,g5,g6)(f(t, _:T2, _:T3, _:T4, _:T5, _:T6)))
+    pp1: T1 => Pretty,
+    pp2: T2 => Pretty,
+    pp3: T3 => Pretty,
+    pp4: T4 => Pretty,
+    pp5: T5 => Pretty,
+    pp6: T6 => Pretty
+  ): Prop = forAllNoShrink(g1, g2, g3, g4, g5, g6)(f)
 
-  /** Universal quantifier for seven explicit generators. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for seven explicit generators. */
   def forAll[T1,T2,T3,T4,T5,T6,T7,P](
     g1: Gen[T1], g2: Gen[T2], g3: Gen[T3], g4: Gen[T4], g5: Gen[T5], g6: Gen[T6], g7: Gen[T7])(
     f: (T1,T2,T3,T4,T5,T6,T7) => P)(implicit
     p: P => Prop,
-    s1: Shrink[T1], pp1: T1 => Pretty,
-    s2: Shrink[T2], pp2: T2 => Pretty,
-    s3: Shrink[T3], pp3: T3 => Pretty,
-    s4: Shrink[T4], pp4: T4 => Pretty,
-    s5: Shrink[T5], pp5: T5 => Pretty,
-    s6: Shrink[T6], pp6: T6 => Pretty,
-    s7: Shrink[T7], pp7: T7 => Pretty
-  ): Prop = forAll(g1)(t => forAll(g2,g3,g4,g5,g6,g7)(f(t, _:T2, _:T3, _:T4, _:T5, _:T6, _:T7)))
+    pp1: T1 => Pretty,
+    pp2: T2 => Pretty,
+    pp3: T3 => Pretty,
+    pp4: T4 => Pretty,
+    pp5: T5 => Pretty,
+    pp6: T6 => Pretty,
+    pp7: T7 => Pretty
+  ): Prop = forAllNoShrink(g1, g2, g3, g4, g5, g6, g7)(f)
 
-  /** Universal quantifier for eight explicit generators. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for eight explicit generators. */
   def forAll[T1,T2,T3,T4,T5,T6,T7,T8,P](
     g1: Gen[T1], g2: Gen[T2], g3: Gen[T3], g4: Gen[T4], g5: Gen[T5], g6: Gen[T6], g7: Gen[T7], g8: Gen[T8])(
     f: (T1,T2,T3,T4,T5,T6,T7,T8) => P)(implicit
     p: P => Prop,
-    s1: Shrink[T1], pp1: T1 => Pretty,
-    s2: Shrink[T2], pp2: T2 => Pretty,
-    s3: Shrink[T3], pp3: T3 => Pretty,
-    s4: Shrink[T4], pp4: T4 => Pretty,
-    s5: Shrink[T5], pp5: T5 => Pretty,
-    s6: Shrink[T6], pp6: T6 => Pretty,
-    s7: Shrink[T7], pp7: T7 => Pretty,
-    s8: Shrink[T8], pp8: T8 => Pretty
-  ): Prop = forAll(g1)(t => forAll(g2,g3,g4,g5,g6,g7,g8)(f(t, _:T2, _:T3, _:T4, _:T5, _:T6, _:T7, _:T8)))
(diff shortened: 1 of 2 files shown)


charleso:
This breaks binary compatibility (that's a great build check to have). I'm not sure what the suggested approach would be. It can actually be made binary compatible if we keep the `Shrink` implicits but ignore them. That _might_ be safer than having to do a whole new major release. I'm happy to discuss whatever approach people think is more appropriate.

dwijnand:
It's probably more acceptable to:

+ keep the implicits in the type signature,
+ don't use them in the implementation,
+ deprecate the methods,
+ Introduce non deprecated methods that doesn't take the Shrink instances,
+ and bump the minor version (x.Y.z).

Or maybe they should be deprecated without changing the behaviour?

charleso:
@dwijnand Thanks for the comment/suggestions. Keeping the implicits is probably the safe way to go. I might wait for a conceptual thumbs up from a maintainer though, I'm worried that it may not be something they're interested in.

I'm not as sure about introducing even more methods though. People can already use `forAllNoShrink` now but `forAll` might appear to be the default except for the shrinking issue. I'm also assuming that we can't have the same `forAll` function with and without an implicit, as that would be ambiguous.

Thanks again.

charleso:
Sorry, that was a clumsy way of saying that I don’t know whether Rickard or others consider this an issue or not. If people are interested I’m very happy to make whatever changes are required. Thanks again.

dwijnand:
Seems I forgot to define the deprecation version in my last change.  I'm logging off, but I can fix it up when I'm back (unless you beat me to it).

charleso:
> Seems I forgot to define the deprecation version in my last change. I'm logging off, but I can fix it up when I'm back (unless you beat me to it).

Fixed. There were (funnily enough) a few warnings around calling the deprecated `forAllNoShrink` that I've fixed up now. Hopefully it's all building correctly now. :)

charleso:
Another person caught out by the forAll behaviour https://github.com/rickynils/scalacheck/issues/443#issuecomment-443321937

rickynils:
@charleso @dwijnand Sorry for my unresponsiveness. I think your approach here makes sense. In one way it makes ScalaCheck less powerful since it removes shrinking that could have been useful in many situations. On the other hand it removes a big source of confusion that a lot of people have stumbled on.

The question is now if we should release a version with unchanged behavior and deprecated methods before actually making this change.

I wonder if we also can remove the hideous `sieve`-thingy from `Gen` when this PR has been merged. It was added as a (poor) attempt at fixing invalid shrinking in some situations.

dwijnand:
> The question is now if we should release a version with unchanged behavior and deprecated methods before actually making this change.

If you prefer, I'm happy if you'd like to do things that way.

1.14.1 with deprecated, same behaviour methods, 1.15.0 with this PR?

> I wonder if we also can remove the hideous `sieve`-thingy from `Gen` when this PR has been merged. It was added as a (poor) attempt at fixing invalid shrinking in some situations.

Sounds good.

src/main/scala/org/scalacheck/Prop.scala
```diff
@@ -804,81 +802,81 @@ object Prop {
 
   }
 
-  /** Universal quantifier for an explicit generator. Shrinks failed arguments
-   *  with the default shrink function for the type */
+  /** Universal quantifier for an explicit generator. */
+  @deprecated("Use the variant without Shrink.  Here only for bincompat.", "1.15.0")
```
ashawley: How does one still use shrinking? In QuickCheck, the default for `forAll` is no shrinking, but there is still support for shrinking with `forAllShrink`?  From what I see so far, that isn't available here.  This reads like turning off shrinking entirely, when the proposal was changing the default.

charleso: > How does one still use shrinking?

It's still there, but _only_ for the variants of the `forAll` function with the implicit `Arbitrary`. I'm proposing to disable it for the `forAll` functions that take `Gen`.

[def forAll[A1,P] (f: A1 => P)(implicit p: P => Prop, a1: Arbitrary[A1], s1: Shrink[A1]): Prop](https://github.com/rickynils/scalacheck/blob/86fdea6aaebbdf8163324a690f3abdd5fa92593d/src/main/scala/org/scalacheck/Prop.scala#L1012)

charleso: >  In QuickCheck, the default for forAll is no shrinking

Do you mean ScalaCheck? QuickCheck has always only [shrunk with Arbitrary](https://github.com/rickynils/scalacheck/blob/136f7df0b5c3b184ec434dbf6ecc4c5977f91ced/src/main/scala/org/scalacheck/Prop.scala#L900) ([if implemented for a specific instance](http://hackage.haskell.org/package/QuickCheck-2.12.4/docs/src/Test.QuickCheck.Arbitrary.html#shrink)) and [_never_ with Gen](http://hackage.haskell.org/package/QuickCheck-2.12.4/docs/src/Test.QuickCheck.Property.html#forAll), which is what I'm proposing to do in this PR.

charleso:
> The question is now if we should release a version with unchanged behavior and deprecated methods before actually making this change.

I don't particularly have an agenda here, other than to stop people getting confused about the shrinking, I'm super happy to do whatever you think is best for ScalaCheck.

I'm just wondering out-loud what upgrading might look like. If we deprecate users upgrading to 1.14.1 will most likely switch every `forAll` to `forAllNoShrink`. But if they ignore the deprecation warning, and upgrade again to 1.15, the warning goes away.

I guess I'm wondering if we shouldn't re-add the non-shrinking _non-deprecated_ `forAll` back in at all (which is what we're doing in this PR at the moment)? If you're using `Gen` you're always encouraged to use `forAllNoShrinking`? And in 1.15 do we perhaps leave `forAll` with `Gen` deprecated but now the behaviour is not to shrink. Purely visually `forAll` is definitely much shorter/nicer than `forAllNoShrink` but perhaps it makes sense to make the split clear given the previous behaviour and history. I'm not sure though. :(

Again, I'd like to help get this change through in whatever form makes sense. I'm actually pretty indecisive and so I'm happy if you have an opinion on how you want the change(s) implemented and I can do the leg work.

src/main/scala/org/scalacheck/Prop.scala
ashawley: Maybe I missed the nuance of the change then.

So you're proposing four cases:

1) `forAll` and explicit `Gen`s with shrinking disabled
3) `forAll` with implicit `Arbitrary`s with shrinking preserved
3) `forAllShrink` and explicit `Gen`s with shrinking preserved
4) `forAllNoShrink` with implicit `Arbitrary`s

And still supporting all the possible arities of each.

Right now, ScalaCheck does:

1) `forAll` and explicit `Gen`s with shrinking
3) `forAll` with implicit `Arbitrary`s with shrinking
3) `forAllNoShrink` with explicit `Gen`s with shrinking
4) `forAllNoShrink` with implicit `Arbitrary`s

charleso:
I've created a PR that just deprecates the `forAll` `Gen` functions as suggested (#444). Let me know if that's a good first step.

src/main/scala/org/scalacheck/Prop.scala
charleso: > So you're proposing four cases:

Yeah something like that. The `forAllShrink` already exists, but takes a manual shrink function. I'm not touching that. I'm only changing the behaviour of `forAll` with `Gen` which is where the problem is. Everything else stays the same.

I suspect what I've proposed at the moment might have to be tweaked based on @rickynils comment and my own counter-question below (which is totally fine).

ashawley: I sense this will be a problem.  There needs to be a `forAllShrink` with an implicit 
 `Shrink`.  I don't entirely grok the ScalaCheck internals, but I believe the existing `forAllShrink` is not of any help here.

charleso:
And a second PR #445 which extends that just to disable shrinking but leave the deprecation warning which is slightly different to what this PR is proposing. Sorry for any confusion around having multiple PRs.

src/main/scala/org/scalacheck/Prop.scala
charleso: > I believe the existing forAllShrink is not of any help here.

As a convenient function for taking X number of `Gen`s with implicit shrinking you're quite right. The current function requires the manual shrinking and only a single Gen.

> There needs to be a forAllShrink with an implicit Shrink

On this I'm afraid I don't agree. I'm curious who is building/using manual `Gen` with a restricted subset of the domain (eg. a subset of `Int`, say between 100 and 200) and still want implicit shrinking that might completely go-outside of that domain (eg. shrink to 0)? How do you know what the global/implicit `Shrink` is going to do and whether it aligns with your custom `Gen` range?

I hope this doesn't sound too lazy, but I don't think I would want to add that missing set of functions unless someone _really_ demanded/wanted it. I would try to convince them first it's a bad/dangerous idea. They can still use `Arbitrary` for their own types with their own `Shrink` implementations. :)

ashawley: > I'm curious who is building/using manual `Gen` with a restricted subset of the domain and still want implicit shrinking?

It's a flexibility issue.  Users could be declaring a `Gen` to `forAllShrink` but still defining their own `Shrink` definition in implicit scope, [which is supported in ScalaCheck](https://github.com/rickynils/scalacheck/blob/f31c3ed/doc/UserGuide.md#test-case-minimisation).

charleso: > Users could be declaring a Gen to forAllShrink but still defining their own Shrink definition in implicit scope,

```scala
implicit def ShrinkMyInt: Shrink[Int] = ... // Shrink towards 10 say

//  This will shrink the second generator incorrectly
forAll(Gen.choose(10, 20), Gen.choose(30, 40)) { ... }
```

In the above example you can only provide _one_ implicit for `Shrink[Int]` in scope, but you really need one for _each_ `Gen` otherwise you're just asking for the same kinds of bugs/errors that this PR is trying to fix. I would argue you really shouldn't be doing anything implicit with `Shrink` if you're providing explicit `Gen`. Now perhaps we might want _explicit_ Gen _and_ Shrink together, which is totally fine...

```scala
forAllShrink[A, B](g1: Gen[A], s1: Shrink[A], g2: Gen[B], s2: Shrink[B])(....)
```

At this point I might suggest making a case class with Gen + Shrink to make it easier to pair/keep them together though.

```scala
case class GenShrink[A](gen: Gen[A], shrink: Shrink[A]) 
```

ashawley: There are ways to use different scopes, but supporting generatorso with shrinking having is a usability issue.  Users need the flexibility to study their failing tests by inserting a more limited generator when they are working on an arbitrary property.

For instance, while writing an arbitrary property:

```scala
    forAllShrink { s: String =>
      ...
```

You may want to study the issue with a specific type of generator without giving up shrinking:

```scala
    forAllShrink(Gen.asciiStr) { s: String =>
      ...
```

charleso: > There are ways to use different scopes

Short of passing in each shrinking manually I don't know how you solve the problem of having two generators of the same type (`Char` in this case):

```scala
forAllShrink(Gen.asciiStr, Gen.alphaNumChar) { (a, b) =>
```

There is no shrinker that will work (correctly) for both those cases, so you would need to pass them in manually:

```scala
forAllShrink
  (Gen.asciiStr, Gen.alphaNumChar)
  (Shrink.asciiStr, Shrink.alphaNumChar)
  { (a, b) =>
```

As soon as you pass in one shrink implicit you also have to pass in the extra `A => Pretty` implicit argument as well, which becomes extremely painful. As I said in my previous comment, if you want to keep supporting this you should pass them in explicitly (no implicitly), and I would be tempted to put them together.

```scala
forAllShrink(
    Gen.asciiStr, Shrink.asciiStr
  , Gen.alphaNumChar, Shrink.alphaNumChar
  ) { (a, b) =>
```

charleso: > There are ways to use different scopes

Sorry we probably should have taken this to Gitter at some point. In summary, just to draw a line under this thread.

1. I don't agree that shrinking should _ever_ be "implicit" for Gen (only Arbitrary).
2. I'm more than happy to add whatever extra functionality that @rickynils would like to see, but given that I disagree with Shrinking being implicit for Gen, I'm not likely to add anything along those lines extra unless he asks.

Let me know if you want to continue this conversation on Gitter or some other medium. :)

ashawley: > I don't agree that shrinking should _ever_ be "implicit" for Gen (only Arbitrary).

This seems like a hard line to draw, but we can set it aside.  Instead, given that`forAll` with `Gen` in ScalaCheck presently supports shrinking, it seems it should still be available to users.
And it can be available without interfering with the `forAll` change proposed here, by having it be named `forAllShrink`.

dwijnand: `forAllYoloShrink`? :trollface:

I have no standing, but I guess it would be fine (and maybe a good migration path) to have a forAll that takes an explicit Gen and an implicit Shrink.

ashawley: There are already is one.  We're talking about when arity is greater than one.

dwijnand: I think we're on the same page; by "a forAll" I meant a variant set of forAll's.

ashawley: Ok, but I'm not sure how useful explicit shrinks would be.  I'd say implicits are a feature here.

joroKr21: FWIW I've been working around this issue for years by using `implicit val noShrink: Shrink[MyType] = Shrink.shrinkAny` (couldn't use `forAllNoShrink` because of scalatest).

I don't think `Gen` with implicit `Shrink`  makes any sense. But I like the idea of combining them. Then the usability issue could be solved by adding a `withShrink(implicit shrink: Shrink[A])` method on `Gen`.

charleso: > Then the usability issue could be solved by adding a withShrink(implicit shrink: Shrink[A]) method on Gen.

I'm curious. I hope you don't mind me asking. What `Gen`s are you constructing/using for your code base, and in particular what is the range of the values? And in what circumstances would you want to summon an `implicit` `Shrink`?

I ask, because I don't really think the implicit part is a good idea (ie `withShrink(shrink: Shrink[A])` would be what I would suggest). But I don't have enough data points to know what people are actually doing with Shrink. I only have my biases. :)

joroKr21: > I ask, because I don't really think the implicit part is a good idea (ie withShrink(shrink: Shrink[A]) would be what I would suggest). But I don't have enough data points to know what people are actually doing with Shrink. I only have my biases. :)

I tend to agree here, I don't think I would be using the implicitness here very often, but sometimes it might come in handy (say I have my own type with `implicit Shrink[MyType]`) and then I do `Gen.listOf(arbitrary[MyType]).withShrink` then shrinking would be automatically propagated to the list generator, but I could still provide explicit shrinkers for other generators.

woparry: Explicit shrinks (ie. forAllShrink with arity > 1) would be extremely useful for me.

My workflow for working around this issue has always been to write Gen and Shrink instances in parallel, and to wrap them (case class AsciiString(string: String), Gen[AsciiString], Shrink[AsciiString]) if there is ambiguity (because that's less annoying than passing in the Pretty instances).

@charleso 's case class GenShrink[A](gen: Gen[A], shrink: Shrink[A]) looks perfect for this, especially .filter could pretty easily operate on both sides.

charleso: @woparry Thanks for the comment, it's good to hear from someone who is using this.

Just curious, any reason you don't provide an implicit `Arbitrary` to make that `Gen` as the global/implicit version? Is it just that `Gen` works with the global `Shrink`, or something else?

ashawley:
> I'm just wondering out-loud what upgrading might look like.

An implicit is being deprecated, so the upgrading is the difficult part of this.  Deprecating a method typically will result in its subsequent removal, or a subsequent removal of some signature of the method.  It's not possible to provide a deprecation time period that gives the old version and simultaneously offers a new alternative without the implicits, since the compiler would be confused about conflicting signatures.  And a `deprecatedName` won't work, since presumably nobody is explicitly passing a `Shrink[T]` to `forAll`.  

As a result, when the implicit is removed the user's code will still compile.  If they miss or ignore the deprecation warning, there will be no indication of the behavior change.  Since most test suites succeed when ScalaCheck is upgraded, there won't be any indication that the shrinking policy changed for `forAll(Gen)`.   They would need to be interacting with a failing property test to notice.  So after deprecating, a `migration` annotation should be added when `forAll(Gen)` is changed to mitigate that problem.

It wasn't a poor choice to use an implicit for shrinking in ScalaCheck.  This is a major behavior change in the library, so it's unsurprising this change requires delicate handling. 

The steps I would propose:

1. When `forAll(Gen)` is deprecated, users should be provided a `forAllShrink(Gen)` method (which doesn't exist for all arities up to 8), rather than `forAllNoShrink`.   During deprecation, `forAllShrink(Gen)` will just be an alias to `forAll(Gen)`.
2.  In the release after deprecation, `forAll(Gen)` would become an alias to `forAllNoShrink(Gen)`, and `forAllShrink(Gen)` would remain.  
3. `forAll(Gen)` would have the `deprecated` annotation changed to a `migration`.
4. The `forAllNoShrink(Gen)` is redundant of `forAll(Gen)`, but should be preserved to maintain parity with `forAllNoShrink(Arbitrary)`.

charleso:
@ashawley 

I've pushed to #444 with changes to the deprecation warning and the multi-arity `forAllShrink` functions as you've suggested. Please let me know if that looks like what you had in mind?

charleso:
@non Thanks again for the detailed response!

I completely agree that a Hedgehog "integrated shrinking" is absolutely 100% the way to go if there's an appetite for it. Even though I'm the author of the Scala Hedgehog library, I strongly believe the wider Scala community would greatly benefit if ScalaCheck gets the same shrinking capability (in the process making Scala Hedgehog mostly redundant).

> Given the new -disableLegacyShrinking flag, do you think there's any other tactical work

I guess I can't help (for the last time I promise) being a broken record and say I still believe disabling shrinking for `forAll` with `Gen` is something we should just do now. I basically see it as a bug fix and bringing it back inline somewhat with the original QuickCheck. It's a source and binary compatible change, and I think it will greatly improve the _default_ interaction with ScalaCheck for _most_ users (as per my [being optimistic comment here](https://github.com/typelevel/scalacheck/pull/522#issuecomment-525925484)). Perhaps it wold make sense to include a `-enableGenShrinking` for users who still really want that ability back (despite the risks).

Anyway, I can see I'm not getting any traction with that argument, and I'm not trying to antagonise anyone over it. I'll close the PR now but I'm happy to re-open or help push something through if there's any subsequent interest. :)

Thanks again for really stepping up to help maintain ScalaCheck, it's already made a huge difference and I'm excited about its future. Good luck!

PR #444, opened 2018-12-07
Deprecate forAll with Gens

charleso:
@rickynils @dwijnand 

Sorry if this gets confusing pretty quickly. This is one option for the first change to ScalaCheck to deprecate the forAll Gen functions, but leave the behaviour. Is it what you had in mind?

EDIT: This is based on the original disabling of Shrinking on #440

### DIFF
--- a/src/main/scala/org/scalacheck/Gen.scala
+++ b/src/main/scala/org/scalacheck/Gen.scala
@@ -163,7 +163,7 @@ sealed abstract class Gen[+T] extends Serializable { self =>
     if (lhs == rhs) Prop.proved(prms) else Prop.falsified(prms)
   }
 
-  def !=[U](g: Gen[U]) = Prop.forAll(this)(r => Prop.forAll(g)(_ != r))
+  def !=[U](g: Gen[U]) = Prop.forAllShrink(this)(r => Prop.forAllShrink(g)(_ != r))
 
   def !==[U](g: Gen[U]) = Prop { prms =>
     // test inequality using a random seed
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -747,6 +747,107 @@ object Prop {
   }
 
   /** Universal quantifier for an explicit generator. Shrinks failed arguments
+   *  with the default shrink function for the type */
+  def forAllShrink[T1,P](
+    g1: Gen[T1])(
+    f: T1 => P)(implicit
+    p: P => Prop,
+    s1: Shrink[T1],
+    pp1: T1 => Pretty
+  ): Prop = forAllShrink[T1,P](g1, shrink[T1] _)(f)
+
+  /** Universal quantifier for two explicit generators. Shrinks failed arguments
+   *  with the default shrink function for the type */
+  def forAllShrink[T1,T2,P](
+    g1: Gen[T1], g2: Gen[T2])(
+    f: (T1,T2) => P)(implicit
+    p: P => Prop,
+    s1: Shrink[T1], pp1: T1 => Pretty,
+    s2: Shrink[T2], pp2: T2 => Pretty
+  ): Prop = forAllShrink(g1)(t => forAllShrink(g2)(f(t, _:T2)))
+
+  /** Universal quantifier for three explicit generators. Shrinks failed arguments
+   *  with the default shrink function for the type */
+  def forAllShrink[T1,T2,T3,P](
+    g1: Gen[T1], g2: Gen[T2], g3: Gen[T3])(
+    f: (T1,T2,T3) => P)(implicit
+    p: P => Prop,
+    s1: Shrink[T1], pp1: T1 => Pretty,
+    s2: Shrink[T2], pp2: T2 => Pretty,
+    s3: Shrink[T3], pp3: T3 => Pretty
+  ): Prop = forAllShrink(g1)(t => forAllShrink(g2,g3)(f(t, _:T2, _:T3)))
+
+  /** Universal quantifier for four explicit generators. Shrinks failed arguments
+   *  with the default shrink function for the type */
+  def forAllShrink[T1,T2,T3,T4,P](
+    g1: Gen[T1], g2: Gen[T2], g3: Gen[T3], g4: Gen[T4])(
+    f: (T1,T2,T3,T4) => P)(implicit
+    p: P => Prop,
+    s1: Shrink[T1], pp1: T1 => Pretty,
+    s2: Shrink[T2], pp2: T2 => Pretty,
+    s3: Shrink[T3], pp3: T3 => Pretty,
+    s4: Shrink[T4], pp4: T4 => Pretty
+  ): Prop = forAllShrink(g1)(t => forAllShrink(g2,g3,g4)(f(t, _:T2, _:T3, _:T4)))
+
+  /** Universal quantifier for five explicit generators. Shrinks failed arguments
+   *  with the default shrink function for the type */
+  def forAllShrink[T1,T2,T3,T4,T5,P](
+    g1: Gen[T1], g2: Gen[T2], g3: Gen[T3], g4: Gen[T4], g5: Gen[T5])(
+    f: (T1,T2,T3,T4,T5) => P)(implicit
+    p: P => Prop,
+    s1: Shrink[T1], pp1: T1 => Pretty,
+    s2: Shrink[T2], pp2: T2 => Pretty,
+    s3: Shrink[T3], pp3: T3 => Pretty,
+    s4: Shrink[T4], pp4: T4 => Pretty,
+    s5: Shrink[T5], pp5: T5 => Pretty
+  ): Prop = forAllShrink(g1)(t => forAllShrink(g2,g3,g4,g5)(f(t, _:T2, _:T3, _:T4, _:T5)))
+
+  /** Universal quantifier for six explicit generators. Shrinks failed arguments
+   *  with the default shrink function for the type */
+  def forAllShrink[T1,T2,T3,T4,T5,T6,P](
+    g1: Gen[T1], g2: Gen[T2], g3: Gen[T3], g4: Gen[T4], g5: Gen[T5], g6: Gen[T6])(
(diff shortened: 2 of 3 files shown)


src/main/scala/org/scalacheck/Prop.scala
```diff
@@ -806,6 +806,7 @@ object Prop {
 
   /** Universal quantifier for an explicit generator. Shrinks failed arguments
    *  with the default shrink function for the type */
+  @deprecated("Use 'forAllNoShrink', shrinking will be disabled for Gen by default in the next release", "1.14.1")
```
charleso: I suck with useful error messages. Suggestions welcome for what might be more appropriate/useful/readable.

ashawley: As I mentioned in #440, it would be better to provide users with `forAllShrink`, since `forAllNoShrink` does not provide shrinking.  

This requires adding all the same arities of `forAllShrink(Gen)` as exist for `forAll(Gen)`.  This would probably end up needing to be released as 1.15.0.

The wording of the deprecation should probably be something like, "Shrink will be disabled when given explicit Gen, use forAllShrink instead", giving:

```
[warn] method forAll in object Prop is deprecated (since 1.15.0): 
[warn] Shrink will be disabled when given explicit Gen, use forAllShrink instead
```

src/test/scala/org/scalacheck/PropSpecification.scala
```diff
@@ -24,7 +24,7 @@ object PropSpecification extends Properties("Prop") {
 
   property("Prop.==> undecided") = forAll { p1: Prop =>
     val g = oneOf(falsified,undecided)
-    forAll(g) { p2 =>
+    forAllNoShrink(g) { p2 =>
```
ashawley: Now that you've defined `forAllShrink(Gen)`, you could use it in the test suite.  But it would be better to just minimize the changes in the PR, by keeping `forAll(Gen)` in the tests.  The deprecation warnings for `forAll(Gen) are non-fatal in the test suite, anyway.

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -163,7 +163,7 @@ sealed abstract class Gen[+T] extends Serializable { self =>
     if (lhs == rhs) Prop.proved(prms) else Prop.falsified(prms)
   }
 
-  def !=[U](g: Gen[U]) = Prop.forAll(this)(r => Prop.forAll(g)(_ != r))
+  def !=[U](g: Gen[U]) = Prop.forAllNoShrink(this)(r => Prop.forAllNoShrink(g)(_ != r))
```
ashawley: Now that you've defined it, you can use `forAllShrink`.  Otherwise, you're changing the semantics of `!=`.

src/main/scala/org/scalacheck/commands/Commands.scala
```diff
@@ -215,7 +215,7 @@ trait Commands {
   final def property(threadCount: Int = 1, maxParComb: Int = 1000000): Prop = {
     val suts = collection.mutable.Map.empty[AnyRef,(State,Option[Sut])]
 
-    Prop.forAll(actions(threadCount, maxParComb)) { as =>
+    Prop.forAllNoShrink(actions(threadCount, maxParComb)) { as =>
```
ashawley: Same here, you can use `forAllShrink`.  Otherwise, you're changing the implementation.  Additionally, there is a `Shrink[Actions]` in scope.

charleso:
Thanks @ashawley, sorry some of that was a little sloppy.

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -163,7 +163,7 @@ sealed abstract class Gen[+T] extends Serializable { self =>
     if (lhs == rhs) Prop.proved(prms) else Prop.falsified(prms)
   }
 
-  def !=[U](g: Gen[U]) = Prop.forAll(this)(r => Prop.forAll(g)(_ != r))
+  def !=[U](g: Gen[U]) = Prop.forAllShrink(this)(r => Prop.forAllNoShrink(g)(_ != r))
```
ashawley: You fixed one, but theres a second one that needs to be `forAllShrink`.

charleso: Apologies, I'm blind.

src/main/scala/org/scalacheck/Prop.scala
```diff
@@ -806,16 +907,18 @@ object Prop {
 
   /** Universal quantifier for an explicit generator. Shrinks failed arguments
    *  with the default shrink function for the type */
+  @deprecated("Use 'forAllNoShrink', shrinking will be disabled for Gen, use forAllShrink instead", "1.14.1")
```
ashawley: The deprecation message still needs tweaking.  As I suggested, it should be

    Shrink will be disabled with explicit Gen, use forAllShrink instead

The version number should be `1.15.0`, not `1.14.1`.

dwijnand:
@ashawley do you have merge and release permissions? I'm wondering if I should weigh in on the changes you're suggesting or not.

ashawley:
I don't, but hopefully you'd weigh in regardless.

src/main/scala/org/scalacheck/Prop.scala
ashawley: I'm not sure how strict ScalaCheck is with semver, but I understand you increment the minor version when there is either new functionality or new deprecations.

ashawley: It's not necessary to suggest `forAllNoShrink`, because that's what `forAll` will effectively become.

ashawley: The tone of the deprecation message could be less commanding.  Users could assume they have to migrate to `forAllShrink`.  It's really their choice.

charleso: @ashawley Thanks, I wasn't paying close enough attention (again). I've just used your full/original suggestion and updated the version number. I've updated the PR. Sorry for the confusion and wasted time.

ashawley:
I don't support deprecating shrinking with an explicit generator.  I'd rather see it fixed, but in the absence of a fix, @charleso has made a valid argument that it should be disabled by default.  My concern of not providing an opt-in for shrinking via `forAllShrink` has been taken in to consideration here.  There's no reason not to move forward with deprecation.

Thanks for doing this!

ashawley:
Concerning Rickard's original question in #440:

> The question is now if we should release a version with unchanged behavior and deprecated methods before actually making this change.

This deprecation warning for`forAll` is likely going to be pretty disruptive for users of ScalaCheck 1.x:

1. New users are going to write their first property test using `forAll` with a generator and be confused by the warning.
2. Users could follow the advice of the deprecation message, "use `forAllShrink` instead", but that is misleading.  The consensus is to not use shrinking, since the default for `forAll` will become equivalent to `forAllNoShrink`.  The suggestion of `forAllShrink` is intended for migrating tests of existing users.
3. Alternatively, users could avoid the warning by using `forAlllNoShrink`, but that's unfortunate because in the next ScalaCheck release (1.16.x, 2.0?) the semantics for `forAll` will become the same as `forAllNoShrink`.
4. Projects with fatal warnings enabled, will need to research how to modify their sbt build and disable fatal warnings in their test suites.  
5. All instances where `forAll` uses an explicit generator in the documentation could be updated to use `forAllNoShrink`, however, like (2) that's a moving target, since the plan is to make `forAll` have shrinking disabled.

We could try to either massage the deprecation message or jam in more information.  Any suggestions to the wording are welcome, but it has diminishing returns.  

I'd wager there needs to be a comprehensive release notes about this.

The release that subsequently removes shrinking should follow closely behind the release that does deprecation.  That would avoid confusing new users from the deprecation warning.

charleso:
Hi @ashawley

Thanks for taking the time to review this PR. My sincere apologies for doing this but I'm going to close the PR. It was originally as a Proof Of Concept and to help communicate one of the options for "fixing" (for some definition) for `forAll (as part of #440 ).

I wont deny it, ScalaCheck is in a tough spot. There is an _extremely commonly used_ function that is behaving in an (arguably) unexpected fashion. By changing the semantics are we breaking compatibility?

Honestly I don't think the there is a Right Answer. I just know that personally I'm going to feel extremely responsible if we merge this PR and, as you 100% correctly point out, "New users are going to write their first property test using `forAll` with a generator and be confused by the warning.".

I'm more than happy if @rickynils, yourself, or another maintainer want to re-open and merge this PR, but I don't want to be seen as the person who was responsible for 1000 deprecation warnings for (in my opinion) a behaviour they didn't realise they were getting.

I still think that #440 is the right approach in this situation. Let's just change the behaviour. The worst that can happen is that users get slightly more unhelpful failure cases, but that are _true_ positives. The alternatives are:

1. They get false positives in the form of confusing errors by leaving broken shrinking
2. They have to refactor _all_ their code to remove deprecation warnings for code that should otherwise by just fine.

I'm really not trying to push you to merge/support #440, I suspect you won't and I'm not trying to have that argument again. I'm just not comfortable with #444 (this PR) and don't want to have my name associated with potentially pissing off lots of people "fixing" something.

Thanks again though, I really mean that, and sorry for dragging out a PR I wasn't really comfortable with in the first place.

ashawley:
Ok, we can resurrect the branch if we need to.   I actually agree that there probably isn't much advantage to this approach of deprecating.  It's probably better to just migrate over in #444.  We still have a lot of the same issues with #444, but it will be less confusing than pushing a release with deprecation warnings.

shawjef3:
I'm wondering what y'all would think about making shrink/noshrink a boolean parameter. Is there an easy way to turn shrinking on and off? Like if I have a test suite and I want to turn shrinking on and off for the whole suite. For instance, perhaps I'm still working on the instance of Shrink for my type and want to turn it on or off in a simple way.

I'm convincing myself this isn't a good idea, since `forAllShrink` would take a `Shrink` parameter, whereas `forAllNoShrink` would not. And so the method with the boolean parameter would still require you to have an instance of `Shrink`. So what I'm thinking of is maybe best if any `shrink` switch is turned on by default.

PR #445, opened 2018-12-07
Disable shrinking for forAll with Gen

charleso:
@rickynils @dwijnand 

Part 2 of #444 after comments on #440. See the second commit in particular. This is one of the alternative options after bumping the minor version to `1.15.x` (I'm assuming I wouldn't do that in this PR though?). I'm not sure whether keeping the deprecated warnings is going to be popular/desired, in which case we can just stick with #440 and close this.

Apologies if having multiple PRs gets confusing.


charleso:
See reasons for closing here: https://github.com/rickynils/scalacheck/pull/444#issuecomment-466302851

PR #446, opened 2018-12-07
Always show seed for failing property check

ryanpbrewster:
This copies the current behavior of `propertyWithSeed(..., None)` into `property(...)`

### DIFF
--- a/src/main/scala/org/scalacheck/Properties.scala
+++ b/src/main/scala/org/scalacheck/Properties.scala
@@ -94,7 +94,8 @@ class Properties(val name: String) {
    */
   sealed class PropertySpecifier() {
     def update(propName: String, p: => Prop) = {
-      props += ((name+"."+propName, Prop.delay(p)))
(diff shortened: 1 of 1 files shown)


ryanpbrewster:
I believe this will address https://github.com/rickynils/scalacheck/issues/400

SethTisue:
rebase?

ryanpbrewster:
done

SethTisue:
is someone knowledgeable enough to review this?

ashawley:
It looks good to me, but @non is the seed expert.

ryanpbrewster:
Note for reviewer: this is basically copy-pasting the implementation from L111, the `PropertyWithSeedSpecifier` class. This class will always show the seed for a failing test case.

((Effectively the only change here is adding a call to `.viewSeed(fullName)` to the end of `Prop.delay(p)`. Let me know if there's anything else I can add here.))

SethTisue:
perhaps add a test that illustrates the change...?

Issue #421, opened 2018-09-11
Overloaded Prop.collect behaves surprisingly with Seq[T] argument


pford19:
**Summary**

`Prop.collect` is overloaded. There is a "simple" version
- `collect(T)(Prop): Prop`

and a "functional" version
- `collect(T=>P)(implicit P=>Prop): T=>Prop`

Using `collect` in properties based on generators that produce `Seq[T]` values (for example `Gen.pick`) can lead to potentially surprising behavior. In the example detailed below, the compiler selects the functional version over the simple version when the argument is of type `Seq[T]` with a `Type mismatch` compiler error. 

Once the issue is understood, there are straightforward workarounds.

Improved documentation in the source file for the two `collect` methods seems a straightforward way to mitigate potentially complex and frustrating diagnosis by your typical developer (assuming she gets as far a clicking through to the Scalacheck sources, or pulling up the equivalent of Intellij's quick documentation).

Removing the overload completely by renaming the functional version is a more ambitious solution.

**Details**

I was using `collect` in a property along these lines

 ` val prop = forAll(Gen.pick(2, (0 to 9)) { ns => ns.size == 2 }`

which I then augmented with `collect`

 ` val prop = forAll(Gen.pick(2, (0 to 9)) { ns => collect(ns) { ns.size == 2 }}`

At which point I get this mysterious (to me, at the time) compiler error 

  `Type mismatch: expected: Int => Prop, actual: Boolean`

Eventually I figured out what was going on, but it took a while, mostly because my eyes start to glaze over when I end up looking at a declaration like: 

  `def collect[T, P](f: T => P)(implicit ev: P => Prop): T => Prop = ...`

The problem is that `pick` is type `Gen[Seq[T]]`, and `Seq[T]` extends `PartialFunction[Int,T]`. This explains the compiler's decision to use `collect(Int=>T)` instead of `collect(Seq[T])`. (At least in part, I wonder why it does complain about ambiguous overloading. But that is a topic for another day.)

Workarounds are relatively simple, such as 

- Cast to `Iterable` (which is not a function type)
 ` val prop = forAll(Gen.pick(2, (0 to 9)) { ns => collect(ns.toIterable) { ns.size == 2 }}`

- Convert to `String` 
 ` val prop = forAll(Gen.pick(2, (0 to 9)) { ns => collect(ns.toString) { ns.size == 2 }}`

- Rewrite to intentionally use functional `collect(Seq[Int] => Boolean)`
[...]

PR #449, opened 2019-01-20
Deprecate overloaded Prop.collect

ashawley:
In support of #421, this PR deprecates the single-argument `Prop.collect`.   The method collects statistics on the _parameter_ to a property test.  It conflicts with the second version of `Prop.collect` that takes two arguments, a value and a property.   The two overloaded methods confuse the Scala compiler for two reasons 1) because the former accepts an implicit second argument and 2) because collections from the standard library typically extend `PartialFunction`.

### DIFF
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -451,6 +451,7 @@ object Prop {
     try { x; false } catch { case e if c.isInstance(e) => true }
 
   /** Collect data for presentation in test report */
+  @deprecated("Use Prop.forAll(t => Prop.collect(t)(...)) instead of Prop.forAll(Prop.collect(t => ...))", "1.15.0")
   def collect[T, P](f: T => P)(implicit ev: P => Prop): T => Prop = t => Prop { prms =>
     val prop = ev(f(t))
     prop(prms).collect(t)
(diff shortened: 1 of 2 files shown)


non:
I think the deprecation message could be slightly improved (I found the name `prop` instead of `f` to be a little bit confusing when I first read it).

Relatedly, I think we can add a law that the rewrite the deprecation warning suggests is guaranteed to work (which could replace the other tests in this PR):

```scala
property("forAll(collect(f)) = forAll(x => collect(x)(f(x)))") =
  forAll { (f: Int => Prop, init: Long) =>
    val seed = rng.Seed(init)
    val prop0 = forAll((x: Int) => collect(x)(f(x)))
    val prop1 = forAll(collect(f))
    def run(p: Prop): Prop.Status =
      p.useSeed("name", seed).apply(Gen.Parameters.default).status
    run(prop0) == run(prop1)
  }
```

What do you think?

ashawley:
> I think the deprecation message could be slightly improved (I found the name `prop` instead of `f` to be a little bit confusing when I first read it).

Well, `prop`, is the argument name for the version of `collect` that won't be deprecated.  That's where that's from.  However, I can understand the confusion.  Maybe it should be more pseudocode-like:

    Use Prop.forAll(t => Prop.collect(t){ ... }) instead of Prop.forAll(Prop.collect(t => ...))

non:
@ashawley Do you want to try to get this in before 1.14.1?

ashawley:
No, this PR needs further review.  I wasn't expecting to include with 1.14.1.

ashawley:
> I think we can add a law that the rewrite the deprecation warning suggests is guaranteed to work (which could replace the other tests in this PR)

That is handy.  It has the advantage of not outputting anything as `collect` usually does.   Otherwise, I'm not sure what the value of such a test is in the long-term.  Admittedly, there isn't much value of the new tests.  They only check that the two syntaxes of `collect` will compile.

larsrh:
Should we go ahead with this for 1.15.0?

non:
👍

PR #455, opened 2019-02-13
Reduce weighting of numeric edge cases

ashawley:
Bill Venners recently gave [a talk](https://www.meetup.com/SF-Scala/events/258488711/) ([video](https://www.youtube.com/watch?v=lKtg-CDVDsI) is online) about property testing with ScalaTest at an SF Scala meetup, and mentioned a defect he found with `Arbitrary[Int]` in ScalaCheck.  He showed in a repl that ScalaCheck gives equal weight to the edge cases  (zero, one, negative one, max, negative max, etc.) as the rest of the numeric range.  This weighting causes edge cases to be 50% of the values in a property test with `Gen.chooseNum`.  This risks running tests repeatedly with the same inputs.

```
+ Gen.chooseNum edge cases: OK, passed 100 tests.
> Collected test data: 
51% non-special
49% special
```

This pull requests reduces the proportion of edge cases to something more reasonable, just 10% of the time.

```
+ Gen.chooseNum edge cases: OK, passed 100 tests.
> Collected test data: 
87% non-special
13% special
```

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -483,6 +483,17 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
   property("negative generators are negative") =
     Prop.forAll(Gen.negNum[Int]) { n => n < 0 }
 
+  property("chooseNum edge cases") = {
+    val minT = -9
+    val maxT = 9
+    val specials = List(minT, -1, 0, 1, maxT);
+    Prop.forAllNoShrink(Gen.chooseNum(minT, maxT)) { x: Int =>
+      Prop.classify(specials.contains(x), "special", "non-special") {
+        true
+      }
+    }
+  }
+
   property("finite duration values are valid") =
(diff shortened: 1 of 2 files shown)


ashawley:
One argument against this change concerns combined probabilities of independent events.

For example, if you generate edge cases 10% of the time for a numeric value instead of 50%, then when you need two numeric values you generate edge cases for both values 1% of the time (0.1 * 0.1) rather than 25% of the time (0.5 * 0.5).  

The rest of the time, 99% of cases, you'd be generating two values where neither are edge cases 81% of the time (0.9 * 0.9) or the two values are an edge case and non-edge case 18% of the time (2 * .9 * .01).

In isolation, repeats of edge cases look bad.  However, "fixing" it has other consequences when generating tuples.  Ideally, ScalaCheck could compensate for [conditional probability](https://en.wikipedia.org/wiki/Conditional_probability) when weighting edge cases for tuples.

ashawley:
In light of my previous comment on Mar-24 and now #485 about `Option`, it seems that changing the weighting of things has unintended consequences.  This could be revisited and taken up at another time.

Issue #360, opened 2017-10-08
How should Properties.overrideParameters work?


ssanj:
This is a question around how Properties.overrideParameters should work and is linked to #233 and #234. I'm seeing slightly different behaviour depending on whether I'm running the Property through SBT __test__ or through SBT __run-main__ (command line) with and without overrides.

Given the following property:

```scala
import org.scalacheck.Properties
import org.scalacheck._

object BasicProps extends Properties("Basic") {

  override def overrideParameters(p: Test.Parameters): Test.Parameters =
     p.withMinSuccessfulTests(1000)

  property("should test basic things") =
    Prop.forAll { n: Int => n != 42 }
}
```

I wanted to run the above property more than a 100 times to provoke a failure. I've overridden __overrideParameters__ and set __withMinSuccessfulTests__ to 1000.

When I run this through SBT with the __test__ command I get the following:

```
[info] + Basic.should test basic things: OK, passed 100 tests.
```

I'm a bit confused as to why the above property wasn't run with a 1000 tests.

When I run this through SBT with __test:run-main__ (command line) I get the following:

```
> test:run-main com.example.validation.extra.BasicProps
[info] Running com.example.validation.extra.BasicProps
+ Basic.should test basic things: OK, passed 1000 tests.
```

In the above case it honours the __withMinSuccessfulTests__ of a 1000.

If I try to override the value of __withMinSuccessfulTests__ from the command line through __run-main__ with _-s_ I get the following:

```
> test:run-main com.example.validation.extra.BasicProps -s 5
[info] Running com.example.validation.extra.BasicProps -s 5
+ Basic.should test basic things: OK, passed 1000 tests.
```

The command line override doesn't seem to work if __withMinSuccessfulTests__ has been set on the property.

If I define the following property without any overrides:

```scala
import org.scalacheck.Properties
import org.scalacheck._

object Basic2Props extends Properties("Basic2") {

  property("should test basic things") =
    Prop.forAll { n: Int => n != 42 }
}
```

Running __test:run-main__ with an override for the number of minimum successful tests, I get:

```
> test:run-main com.example.validation.extra.Basic2Props -s 5
[info] Running com.example.validation.extra.Basic2Props -s 5
+ Basic2.should test basic things: OK, passed 5 tests.
```

It looks like the default number of minimum successful tests are overridden in this case.

Running with SBT __test__ uses the defaults:

```
[info] + Basic2.should test basic things: OK, passed 100 tests.
```

[...]

Issue #289, opened 2016-10-27
Properties.overrideParameters doesn't work as expected


alexandrnikitin:
`Properties.overrideParameters` method gets called but doesn't override parameters.
A piece of code to reproduce:

``` scala
import org.scalacheck.Test.Parameters
import org.scalacheck._

object StringSpecification extends Properties("String") {
  import Prop.forAll

  override def overrideParameters(p: Parameters): Parameters = {
    println("called")
    p.withMinSuccessfulTests(1) // this doesn't affect number of iterations
  }

  property("startsWith") = forAll { (a: String, b: String) =>
    println("iteration")
    (a+b).startsWith(a)
  }

}
```

Related to #234 but it didn't get called before this PR.

ljwagerfield:
Has anyone found a solution / workaround to this?

ashawley:
@ljwagerfield See #360 where this was more recently picked up

ashawley:
Closing this issue.  A fix for overriding the parameters of properties was merged in #522.

PR #463, opened 2019-03-24
Allow Properties.overrideParameters to work correctly.

### LINKED ISSUE #221: Properties.overrideParameters doesn't do anything?
Maybe I'm missing something but the `Properties.overrideParameters` method doesn't seem to actually do anything. The comment says it's supposed to override the test parameters for current Properties class but all it seems to do is return the value it's passed and it doesn't have any impact on the test parameters used by the tests.

``` scala
/**
 * Changes to the test parameters that are specific to this class.
 * Can be used to set custom parameter values for this test.
 */
def overrideParameters(p: Test.Parameters): Test.Parameters = p
```

non:
Currently when run from SBT (i.e. ScalaCheckFramework),
the overrideParameters method on Properties doesn't
successfully override any parameters. This was because
of the way the framework handles overrides coming from
build configuration -- the framework unconditionally
writes all parameter values, even when not explicitly
set in the build.

This commit reverses the order that these overrides are
applied, so that build overrides (which are coarser) can
be overridden by overrides in specific properties.

Fixes #221, #289, and #360.

### DIFF
--- a/src/main/scala/org/scalacheck/ScalaCheckFramework.scala
+++ b/src/main/scala/org/scalacheck/ScalaCheckFramework.scala
@@ -29,11 +29,7 @@ private abstract class ScalaCheckRunner extends Runner {
 
   def deserializeTask(task: String, deserializer: String => TaskDef): BaseTask = {
     val taskDef = deserializer(task)
-    val countTestSelectors = taskDef.selectors.toSeq.count {
-      case _:TestSelector => true
-      case _ => false
-    }
-    if (countTestSelectors == 0) rootTask(taskDef)
+    if (taskDef.selectors.isEmpty) rootTask(taskDef)
     else checkPropTask(taskDef, single = true)
   }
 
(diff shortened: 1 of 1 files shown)


ashawley:
In `ScalaCheckFramework.scala`, a comment was added:

> apply global parameters first, then allow properties to
> override them. the other order does not work because
> applyCmdParams will unconditionally set parameters to default
> values, even when they were overridden.

It's too bad we can't have both...override parameters in a class but be able to still override them from the command-line.  

That would also confuse the semantics of what an "override" does.

ashawley:
Maybe add a note that this is the case in the doc comment for `Properties.overrideParameters`.

Write now the doc reads:

> Changes to the test parameters that are specific to this class.
> Can be used to set custom parameter values for this test.

Should be something along the lines of:

> Override the default test parameters for this test.
> 
> Parameters that are set in this override will be ignored when
> set on the command-line.

non:
@ashawley I agree the situation isn't ideal. I may try to work a bit harder to get things working as they were intended, but it's proved a tough nut to crack. Otherwise I agree that I'll need to change the documentation as you say.

non:
All that aside, I think this order is probably better in the long run. For example, `minSuccessfulTests` defaults to `100`. If an author overrides a specific test (A) to set that to `10` (because it's slow), and then later decides that `200` is a better baseline for their tests, we have two options:

 - Require A to pass 200 times (very slow)
 - Still only run A 10 times (requiring the author to update that test to change the # of runs)

Personally, I would consider a local override (in code) to be more specific than a global override (on the command-line), since in the other order there's no way to allow different tests to use different values.

All this aside, existing ScalaCheck docs say it should work the other way, so I'll see if I can get that working. I just thought it was worth making the case that this would possibly be a better design in the long run.

ashawley:
There are other parameters besides just test counts, where you could make the same case or a different case.  I tend to agree, though.  If it is set in the class with `overrideParameters`, then it is fine to ignore even an explicit ScalaCheck argument that is given through sbt. 

It seems you've made the simplest fix, and is a good first pass at making it work.  It hasn't worked since ScalaCheck 1.13.0 came out a few years ago.  In my eyes, fixing it like this shouldn't close other avenues for changing it later.

PR #470, opened 2019-05-08
Add Generators For Hex Char And String Values

isomarcte:


isomarcte:
Is there anything I need to do to get this merged?

ashawley:
This seems like a nice first contribution.  Making this generally useful might be a little more complicated, though.  I wonder how much it is needed.  What context were you using it for?

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -766,6 +766,15 @@ object Gen extends GenArities with GenVersionSpecific {
...
 
+  /** Generates a character that can represent a valid hexadecimal digit. This
+    * includes both upper and lower case values.
+    */
+  def hexChar: Gen[Char] = Gen.oneOf(
+    Gen.choose(48.toChar, 57.toChar),  // 0-9
+    Gen.choose(97.toChar, 102.toChar), // a-f
+    Gen.choose(65.toChar, 70.toChar)   // A-F
+  )
+
```
ashawley: This could be:

```scala
  def hexChar: Gen[Char] =  Gen.frequency(
    10 -> Gen.oneOf('0' to '9'),
     3 -> Gen.oneOf('a' to 'f'),
     3 -> Gen.oneOf('A' to 'F'))
```

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -803,6 +812,12 @@ object Gen extends GenArities with GenVersionSpecific {
   def asciiPrintableStr: Gen[String] =
     listOf(asciiPrintableChar).map(_.mkString)
 
+  /** Generates a string that can represent a valid hexadecimal digit. This
+    * includes both upper and lower case values.
+    */
+  def hexStr: Gen[String] =
+    listOf(hexChar).map(_.mkString)
+
```
ashawley: This produces a lengthy string by default (up to 100?).  That works for your test with `BigInteger`, but may not for when people want byte, int or long from hex?

I guess you've copied the definition of `Gen.numStr`.  I guess it has these same sort of problems.

src/main/scala/org/scalacheck/Gen.scala
isomarcte: I agree. I was trying to be consistent with the rest of the file.

src/main/scala/org/scalacheck/Gen.scala
isomarcte: Again, I agree. I did this for consistency.  That being said, I'm not too worried about this. If people want to create a `String` which represents a bounded type, such as `i32`, then it would be consistent with current idioms to just map on this generator.

isomarcte:
@ashawley parsing or emitting is, in my experience, a pretty common task in CS domains. My specific use case was emitting values for an API which required hex numbers.

non:
@isomarcte Hey, I'm going through the PRs one more time before 1.14.1 and came back to this.

I'm not necessarily wedded to @ashawley's proposed definition, but I do think it would be best if we had equal probability of generating each of the 16 hex digits (right now the last 6 digits are twice as likely as the first 10). You could fix this the way @ashawley suggested, or by another method (I sort of like `Gen.choose(Gen.choose("0123456789abcdef"), Gen.choose("0123456789ABCDEF"))` due to how explicit it is).

I think your rationale for generating large strings in `hexStr` is correct. If you're willing to fix the frequencies of `hexChar` I'm happy to merge this before releasing 1.14.1.

I'm sure it's annoying to get this feedback after so long, but

isomarcte:
@non it's not annoying at all. I completely agree about your reasoning with the probability being equally distributed across the domain. I'll update it. Your proposed definition looks great to me.

isomarcte:
@non updated

non:
Looks good! 👍

PR #472, opened 2019-05-18
Update to 2.13.0-RC2

travisbrown:
Also updated sbt, the sbt-scala-native patch version, and 2.12 since I was here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ notifications:
 scala:
   - 2.10.7
   - 2.11.12
-  - 2.12.6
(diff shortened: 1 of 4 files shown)


travisbrown:
Looks like the failures are 2.13.0-RC1 only and just MiMa not finding a previous release.

dwijnand:
There's an alternative PR for this, #473, which targets the 1.14.0_sonatype branch.

travisbrown:
Ah, great, thanks, @dwijnand.

PR #476, opened 2019-05-27
Make Pretty.break tail recursive

GytisZ:
Hi, we have a rather large test, that would trigger stack overflow from `Prettybreak`. This should fix it.

Tried to keep to similar style. The only functional change is the added requirement for `lead.length`.

No tests submitted, since it seems `util.Pretty` is not tested a lot. I'm leaning towards adding them, but don't want change too much, so waiting for yay/nay whether to do it.

### DIFF
--- a/src/main/scala/org/scalacheck/util/Pretty.scala
+++ b/src/main/scala/org/scalacheck/util/Pretty.scala
@@ -51,9 +51,23 @@ object Pretty {
     if(s.length >= length) s
     else s + List.fill(length-s.length)(c).mkString
 
-  def break(s: String, lead: String, length: Int): String =
-    if(s.length <= length || length <= 0) s
-    else s.substring(0, length) / break(lead+s.substring(length), lead, length)
+  def break(s: String, lead: String, length: Int): String = {
+    require(lead.length < length, "`lead.length` must be shorter than line `length`")
+
+    @annotation.tailrec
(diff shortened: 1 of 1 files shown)


ashawley:
> Hi, we have a rather large test, that would trigger stack overflow from `Prettybreak`.

Do you have an example of a failing test?

GytisZ:
Not in a form I could share easily. In short it's a `org.scalacheck.commands.Commands` test that tests some stateful storage and generates lists of messages to write / read.

A bit longer story is that while the commands themselves are short
```
case class WriteMessages(content: String, numberOfCopies: Int)

val exampleCommand = WriteMessages('foobar', 5)
```

all the messages are kept in `State` written out
```
val s: State = ???
s.currentMessages = List('foobar', 'foobar', 'foobar', 'foobar', 'foobar')
```
and thus a lot longer when written out as a `String` .

scalacheck is writing out these states upon test failure, which then causes `StackOverflow`

This might be a rare occurrence, and maybe our tests are badly designed, but regardless, I think it shouldn't end in a StackOverflow from the framework.

non:
I ended up writing some tests to nail down the break behavior, and ended up with a slightly different stack-safe refactor:

https://github.com/rickynils/scalacheck/pull/495

non:
Closing this because we merged https://github.com/rickynils/scalacheck/pull/495.

@GytisZ thanks for your PR and I'm sorry it didn't get more attention sooner.

PR #479, opened 2019-06-07
Scala 2.13.0

joan38:


Philippus:
`.travis.yml` should be updated as well.

ashawley:
In order for MiMa to pass, 2.13.0 actually has to be published, which see #480.

joan38:
@ashawley What do you mean by published?
https://github.com/scala/scala/releases/tag/v2.13.0
https://mvnrepository.com/artifact/org.scala-lang/scala-library/2.13.0

ashawley:
Sorry, I should have said "ScalaCheck 1.14.0 artifacts for 2.13.0 have to be published".

rossabaker:
It can be temporarily disabled for a particular Scala version, though someone has to follow up and remove the workaround after the release:

```scala
  mimaPreviousArtifacts := (CrossVersion.binaryScalaVersion(scalaVersion.value) match {
    case "2.13" => Set.empty
    case _ => Set("org.scalacheck" %% "scalacheck" % "1.14.0")
  }),
```

Could also add a check on the version number so the workaround self-destructs until it's removed.

joan38:
Alright this is green.
A big chunk of the Scala ecosystem is held by this upgrade. I guess we should move forward quickly?

joan38:
Just talked on https://github.com/scalatest/scalatest/issues/1601 and it seems that this PR is hanging ScalaTest which pretty much the whole Scala community depends on.

ashawley:
This PR is important for moving `master` to 2.13, but publishing 2.13 will be happening from the `1.14_sonatype` branch if the past is any indication of the future, which see #481.  Rickard published RC2 and RC3 pretty quick, so he may just be unavailable.

So the MiMa exception for 2.13 that was suggested here, doesn't need to be done here for `master`.  Coincidentally, it's part of the `1.14_sonatype` branch so that it can publish.

PR #481, opened 2019-06-10
Publish Scala 2.13.0 for 1.14.0

ashawley:
I verified that the following works locally from sbt:

```
> ++2.13.0
> jvm/package
> js/package
```

That will publish Scala 2.13 and Scalajs 0.6.x for 2.13.

Publishing Scalajs for 1.0.0-M8 for 2.12 and 2.13, also works.  First, start sbt with:

    $ env SCALAJS_VERSION=1.0.0-M8 sbt

Then:

```
> ++2.12.6
> js/package
> ++2.13.0
> js/package
```

May require #475 for publish to work, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ scala:
   - 2.10.7
   - 2.11.12
   - 2.12.6
-  - 2.13.0-RC2
+  - 2.13.0
 jdk:
   - oraclejdk8
 env:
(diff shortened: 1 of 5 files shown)


rickynils:
I have now published 2.13 builds of ScalaCheck 1.14.0.

Thanks @ashawley

PR #482, opened 2019-06-10
MiMa setting for Scalajs milestones

ashawley:
A cherry-picked version of #475 to master branch.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -90,7 +90,13 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
   // don't use fatal warnings in tests
   scalacOptions in Test ~= (_ filterNot (_ == "-Xfatal-warnings")),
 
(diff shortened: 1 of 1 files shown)


non:
👍

non:
One question about all of this: most projects explicitly set a scala JS version in `project/plugins.sbt` which is used in all cases. I noticed ScalaCheck seems to have a set up where particular users can change the version with an environment variable.

Is this desirable? Maybe I'm out of touch with modern ScalaJS practices but it seems to me like it would be better to make the build/release process _more_ standardized rather than requiring extra environment variables to be set.

(I'm planning to merge this either way, I'm just going to be refactoring the build to make it possible for me to do a ScalaCheck release and I want to figure out if this environment variable stuff is something I can tear out or not.)

ashawley:
Yeah, I agree it's clunky, but I'm not aware of another way.

non:
I'm going to merge this as-is (but it might end up getting changed when I refactor the build).

Thanks!

PR #484, opened 2019-07-06
Add Shrink.suchThat

koterpillar:
When using `Gen.suchThat` or generators that maintain invariants by construction, shrinking can break those invariants. Add `Shrink.suchThat` to ensure no invalid values are shrunk to.

### DIFF
--- a/doc/UserGuide.md
+++ b/doc/UserGuide.md
@@ -461,6 +461,13 @@ enough values, and it might report a property test as undecided. The
 half of the generated numbers, but one has to be careful when using the
 `suchThat` operator.
 
+Note that if a property fails on a value generated through `suchThat`, and is
+later shrunk (see [test case minimisation](#test-case-minimisation) below, the
+value ultimately reported as failing might not satisfy the condition given to
+`suchThat`, although it doesn't change the fact that there _exists_ a failing
+case that does. To avoid confusion, the corresponding shrink for the type can
+use `suchThat` method too.
+
 #### Generating Containers
 
 There is a special generator, `Gen.containerOf`, that generates containers such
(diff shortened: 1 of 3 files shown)


doc/UserGuide.md
```diff
@@ -827,6 +834,29 @@ When implementing a shrinking method, one has to be careful to only return
...
+  implicitly[Shrink[List[Int]]].suchThat(_.length % 2 == 0)
+```
+
+Note that if a property fails on a value generated through `suchThat`, and is
+later shrunk (see [test case minimisation](#test-case-minimisation) below),
+the value ultimately reported as failing might not satisfy the condition given
+to `suchThat`, although it doesn't change the fact that there _exists_ a
+failing case that does. To avoid confusion, the corresponding shrink for the
+type can use `suchThat` method too.
+
```
ashawley: I think there's an extra copy paste here?  Feel free to edit your intermediate commits witih rebase.  Something strange is happening when you're trying to fix the paren for the "see below".

ashawley:
This is a nice contribution.  I don't have any objection.  It helps improve a correspondence in the APIs of generators and shrinks.  If we can just fix the strange commits to the README, this looks good to me.

doc/UserGuide.md
koterpillar: I originally duplicated it because it made sense both here and above, but now I've incorporated some changes into the paragraph above the example instead.

doc/UserGuide.md
```diff
@@ -827,6 +834,23 @@ When implementing a shrinking method, one has to be careful to only return
...
+```scala
+/** Generate lists of an even length */
+val genEvenList: Gen[List[Int]] = Gen.sized { size =>
+  Gen.listOfN(size * 2, arbitrary[Int])
+}
+
+/** Shrink a list, maintaining even length */
+val shrinkEvenList: Shrink[List[Int]] =
+  implicitly[Shrink[List[Int]]].suchThat(_.length % 2 == 0)
```
ashawley: Is it possible to make `shrinkEvenList` implicit?  Even if I make it lazy, I get an infinite loop when I use it in a property test.

In your test, you use the explicit shrink version of `forAll`.  Unfortunately, that only works for single arity `forAll`.

ashawley: Please review #644 when you get a chance, thanks.

Issue #485, opened 2019-07-10
Circe is broken in community build


ashawley:
There isn't a lot in master, since 1.14.0 was released, but the community build uncovered a test suite error in circe that could be ScalaCheck related:

```
[info] JsonCodecMacrosSuite:
[info] io.circe.generic.JsonCodecMacrosSuite *** ABORTED ***
[info]   java.lang.StackOverflowError:
[info]   at io.circe.generic.jsoncodecmacrossuiteaux.SelfRecursiveWithOption$.$anonfun$arbitrarySelfRecursiveWithOption$1(JsonCodecMacrosSuite.scala:154)
[info]   at org.scalacheck.ArbitraryLowPriority$$anon$1.arbitrary$lzycompute(Arbitrary.scala:75)
[info]   at org.scalacheck.ArbitraryLowPriority$$anon$1.arbitrary(Arbitrary.scala:75)
[info]   at org.scalacheck.ArbitraryLowPriority.$anonfun$arbOption$1(Arbitrary.scala:339)
[info]   at org.scalacheck.ArbitraryLowPriority$$anon$1.arbitrary$lzycompute(Arbitrary.scala:75)
[info]   at org.scalacheck.ArbitraryLowPriority$$anon$1.arbitrary(Arbitrary.scala:75)
[info]   at org.scalacheck.ArbitraryLowPriority.arbitrary(Arbitrary.scala:79)
[info]   at org.scalacheck.ArbitraryLowPriority.arbitrary$(Arbitrary.scala:79)
[info]   at org.scalacheck.Arbitrary$.arbitrary(Arbitrary.scala:62)
[info]   at io.circe.generic.jsoncodecmacrossuiteaux.SelfRecursiveWithOption$.atDepth(JsonCodecMacrosSuite.scala:147)
[info]   ...
[info] SemiautoDerivedSuite:
[info] io.circe.generic.JsonCodecMacrosSuite *** ABORTED ***
[info]   java.lang.StackOverflowError:
[info]   at io.circe.generic.jsoncodecmacrossuiteaux.SelfRecursiveWithOption$.$anonfun$arbitrarySelfRecursiveWithOption$1(JsonCodecMacrosSuite.scala:154)
[info]   at org.scalacheck.ArbitraryLowPriority$$anon$1.arbitrary$lzycompute(Arbitrary.scala:75)
[info]   at org.scalacheck.ArbitraryLowPriority$$anon$1.arbitrary(Arbitrary.scala:75)
[...]

PR #486, opened 2019-07-17
Add test for Arbitrary recursive Option

ashawley:
This is something that came up in #485 as a result of changing `Arbitrary.arbitrary[Option]` for #401.  It seems that tests that had a generator of `X(Option[X])` with faulty recursion were previously able to get away with it.  Presumably, this was because `Arbitrary.arbOption` was more biased to `None`.


ashawley:
Succeeded by #487.

PR #491, opened 2019-08-06
Clean up warnings and scalac options.

non:
This removes -deprecation warnings for 2.13 (since there are over 100
of them). It also tries to make it a bit easier to see which options
are used for which versions. Finally it removes an unused import.

I did this as an isolated PR because it's small and easy to review.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -62,23 +62,40 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
 
   javacOptions += "-Xmx1024M",
 
-  scalacOptions ++= Seq(
-    "-deprecation",
-    "-encoding", "UTF-8",
-    "-feature",
-    "-unchecked",
-    "-Xfuture",
-    "-Ywarn-dead-code",
-    "-Ywarn-numeric-widen") ++ {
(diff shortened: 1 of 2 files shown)


non:
cc @dwijnand @SethTisue

I'm going to tag you all on these PRs that are primarily build-related. Let me know if anything jumps out at you.

ashawley:
I worry this won't work for adding Dotty to the build.  After #423 is merged, I predict people will relatively quickly request that Dotty be added as a compiler so there's no backsliding.

non:
@ashawley I think the strategy will be easily to modify to support dotty. For example:

```scala
mk(211 to 214(...),
mk(211 to 213)(...),
mk(212 to 300)(...),
mk(214 to 301)(...))
```

In the worst case (where the ranges need to be disjoint) we can widen that type to `IndexedSeq[Int]`, e.g. `(211 to 213) :+ 300` (skipping `214`).

ashawley:
Dotty versions are of the scheme "0.17.0-RC1", e.g. https://contributors.scala-lang.org/t/announcing-dotty-0-17-0-rc1/3556

smarter:
Yeah, this will need to be changed for Dotty support. The usual way to do this sort of things is to pattern match on `CrossVersion.partialVersion(scalaVersion.value)`: https://github.com/scala/scala-module-dependency-sample#depend-on-scala-modules-like-a-pro

PR #468, opened 2019-04-17
Fix shrinking of failed test cases in Commands

nrinaudo:
This should fix #447.

Remaining questions:
* why does this fix the issue?!
* if I wanted to write a test for this, how would I go about it?

PR #495, opened 2019-08-07
Stack-safe implementation of Pretty.break.

non:
The motivation for this is #476. I wanted to ensure we had stronger
tests for Pretty.break before refactoring it, and also thought that
this implementation was a little bit nicer.

This PR includes a test for stack-safety, a test to ensure we never
create lines longer than the maximum allowed, and that we can
reassemble the original string from the output. The old code passed
all of these (apart from the stack safety test which it failed), which
helps give confidence and that the new implementation is equivalent.

It also includes a Scaladoc method to explain what the method does.

Thanks to @GytisZ for reporting the issue and providing an initial PR.

### DIFF
--- a/src/main/scala/org/scalacheck/util/Pretty.scala
+++ b/src/main/scala/org/scalacheck/util/Pretty.scala
@@ -13,9 +13,8 @@ import language.implicitConversions
 
 import org.scalacheck.Prop.Arg
 import org.scalacheck.Test
-
-import math.round
-
+import scala.annotation.tailrec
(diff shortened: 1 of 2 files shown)


ashawley:
I'm pretty confident that #468 in 2.12 is the reason this weakness suddenly showed up for `Pretty.break`.  

However, there's no reason not to fix the possible stack overflow.  Toying with it in #477, I also found it to be really slow, so it's good to optimize, as well, so users don't have to wait long to see their output.

PR #503, opened 2019-08-08
Cut out a bunch of Travis builds to see if things are better.

non:
It's possible this is too aggressive (or not aggressive enough) but I
wanted to see how much faster and more reliable things would be with
these settings.


non:
Well, things are still super flaky (and it takes awhile to get scheduled). On the up-side, we only had 14 jobs and the longest took only 8 minutes, which is an improvement.

non:
@ashawley What do you think? It might be nice to do something like this (at least until Travis gets less flaky). For the last couple days I think we've all had to baby-sit these Travis builds and click restart on the failed jobs which is annoying.

ashawley:
Yeah, makes sense.  Looks like it cuts the build times in half.

SethTisue:
I don't know the history of `SBT_PARALLEL`. would it make sense to use `include:` to add a *single* job with `SBT_PARALLEL=false`?

and if that *doesn't* make sense, then I wonder if it might make more sense to eliminate `SBT_PARALLEL` entirely

PR #510, opened 2019-08-14
Remove Cogen.apply with dummy implicit

ashawley:
The change to `Cogen.apply` was flagged as a possible binary compatibility issue in #508.

The basis for this change in #403 was because of some ambiguity in Scala 2.12 with `-Xexperimental`.  

However, the dummy implicit didn't ship with 1.14.0, and its absence didn't seem to be an issue.  It seems downstream projects using `Cogen` haven't had an issue after upgrading to 1.14.0: cats, cats-effect, kittens, enumeratum, scalaz.


non:
Yeah, I think as long as we never actually released this I'm 100% fine with reverting it.

Thanks for investigating the situation!

PR #516, opened 2019-08-16
Use Travis snapshot in examples

ashawley:
Use last released version, or `TRAVIS_COMMIT` version if set.


non:
So I guess the idea here is that we can update examples as soon as the code changes?

I'm a bit nervous that someone using the examples to get started will end up depending on a random snapshot version of ScalaCheck, but maybe that's not a realistic concern.

ashawley:
Yeah, I was trying to make the examples more useful for testing purposes, but you're right it does muddle up the code for users.

PR #519, opened 2019-08-20
Remove invalid filter from MiMa

ashawley:
I predict this was a filter that was added only because the build was incorrectly comparing Scalajs with regular Scala.  Once the configuration was fixed in #482, the filter was no longer necessary.   This resets the Mima settings to empty, except for disabling the MiMa 0.5 signature checking rule.


non:
👍

Issue #443, opened 2018-11-30
Shrinking shrinks a List required to be nonEmpty to Nil


letalvoj:
Let's have a generator of a data structure which contains a *non empty* list. Like
```scala
Gen.nonEmptyListOf(Gen.choose(0,10))
```

When shrinking kicks in, what happens is that the list gets shortened - well that's an obvious way to reduce the input, I agree. Yet if the error is caused by a different parameter, than the shrinking does not stop and shrinks the list all the way to `Nil`.

That kind of sucks, because there is definitely a reason for the list to be non empty!  In my case it causes an `empty.head` error which is completely misleading and obscures the actual problem.

IMO the `Shrinker` should respect the domain of the values specified by the defined `Generator`.

charleso:
Unfortunately in ScalaCheck (and the original QuickCheck) shrinking has to be done separately and in addition to generation. It doesn’t compose.

This problem is made slightly more annoying in ScalaCheck because of the combination of restricted Gens with the global/implicit Shrinker in forAll. I would argue it’s a bug and have opened a PR to remove that particular behaviour (but don’t know if it will be accepted). So you basically wouldn’t get shrinking when using Gen, which is why you’re getting an error, which is a reoccurring problem from what I have observed.

https://github.com/rickynils/scalacheck/pull/440

I highly recommend checking out a fairly new library Hedgehog which effectively fixes/solves this problem by integrating shrinking with Gen. There’s a great talk by the creator. 

https://m.youtube.com/watch?v=AIv_9T0xKEo

Shameless plug warning. There is also the Scala version which I maintain.

https://github.com/hedgehogqa/scala-hedgehog

letalvoj:
Thanks. I've already solved the issue, yet I did not know that there was a
meaningful alternative (hedgehog).

It came to mind that combining Gen with Shrink might be out of the scope of
this project, yet I failed to find a related issue so I filled one.

Interestingly it first emerged when I used nonEmptyListOf ... Filtering out
empty lists manually caused the error rate decrease significantly - such
that the tests stopped failing for a few consecutive runs - which made me
thinking that there was error in the nonEmpty variant of the listOf
function actually.

Anyway, this is a serious design flaw and therefore shrinking should be
disabled by default - to force the user to learn about it first. If you do
not agree feel free to close the issue.

Issue #432, opened 2018-10-03
Shrinking an argument will regenerate and reshrink subsequent arguments


steinybot:
**Expected**

When a property fails then each argument would only be shrunk and new arguments would not be generated.

**Actual**

The last argument is shrunk first. When that argument can not be shrunk any more then it moves onto the preceding argument. The preceding argument is shrunk one step and then the remaining arguments are regenerated. These regenerated arguments must then be shrunk all over again. This continues repeatedly until eventually the first argument has been shrunk.

If you combine this with a generator that filters out values, have arguments that shrink slowly higher up in the argument order, or get unlucky and the first argument is a large value then the shrinking can take an extremely long time.

**Steps to reproduce**

```scala
scala> import org.scalacheck.Prop.forAll
import org.scalacheck.Prop.forAll

scala> import org.scalacheck.Arbitrary
import org.scalacheck.Arbitrary

scala> case class Num(x: Int)
defined class Num

scala> implicit val arbNum = Arbitrary {
     |   Arbitrary.arbitrary[Int].map(Num)
     | }
arbNum: org.scalacheck.Arbitrary[Num] = org.scalacheck.ArbitraryLowPriority$$anon$1@2db6ad21

scala> val propShrink = forAll { (a: Int, b: Num, c: Int, d: Num) =>
     | println("----------")
     | println(s"a: $a")
     | println(s"b: $b")
     | println(s"c: $c")
     | println(s"d: $d")
     | false
     | }
propShrink: org.scalacheck.Prop = Prop

scala> propShrink.check
----------
a: -775860897
b: Num(-1)
c: 1417213460
d: Num(-1)
----------
a: -775860897
b: Num(-1)
c: 708606730
d: Num(-2147483648)
----------
a: -775860897
b: Num(-1)
c: 354303365
d: Num(1)
...
----------
a: -775860897
b: Num(-1)
c: 2
d: Num(0)
----------
a: -775860897
b: Num(-1)
c: 1
d: Num(300617567)
----------
a: -775860897
b: Num(-1)
c: 0
d: Num(-1)
----------
a: -387930448
b: Num(-1963866452)
c: 587290664
d: Num(1)
----------
a: -387930448
b: Num(-1963866452)
c: 293645332
d: Num(1784505235)
----------
a: -387930448
b: Num(-1963866452)
c: 146822666
d: Num(-1)
```

You may need to run the check a few times until the first argument is large enough to reproduce the problem.

[...]

Issue #347, opened 2017-07-23
Gen.choose in 1.13.4 (or 1.13.x?) ignores its range


Kevin-Lee:
`org.scalacheck.Gen.choose` in 1.13.4 (or maybe 1.13.x) seems to ignore its range when it gets a failed test.
For instance, when the following check fails, I think the `ARG_0` should be 10 but it isn't. 
```scala
property("a number should be between 1 and 10") =
  forAll(Gen.choose(1, 10)) { x =>
    println(s"x: $x")
    x > 0 && x < 10 // deliberately made it < 10 so that it fails when x is 10
  }
```
In my case, it gives me 0 which shouldn't be given by `Gen.choose(1, 10)`.

* Result:
```
x: 8
x: 7
x: 10
x: 5
x: -5
x: -2
x: -1
x: 0
! GenChooseSpec.: Falsified after 2 passed tests.
> ARG_0: 0
> ARG_0_ORIGINAL: 10
Found 1 failing properties.
```
As you can see, it even supplies a negative number.
I can see that an old version like 1.12.6 doesn't have this issue.

Kevin-Lee:
Because of this issue, all other gen methods using `choose` (e.g. `numChar`, `alphaUpperChar`, `alphaLowerChar`, `alphaChar`, `alphaNumChar`, `numStr`, `alphaUpperStr`, `alphaLowerStr`, `alphaStr`, `alphaNumStr`, and so on) are broken as well.
e.g.)
`Gen.alphaStr` sometimes gives me `Int` values or some symbols like `+`.

I can't use 1.12.x versions because there is another issue in there (more precisely when using it with Scalatest).

I'm just wondering how this could happen. The test framework was released without testing itself? 😞

Kevin-Lee:
I just saw #317. If it's an already known issue, how come I don't have this in 1.12.x? Does that mean I was just lucky when I was using 1.12.x? 🤔

rickynils:
@Kevin-Lee I guess you were lucky, but there might have been changes that affected the behavior.

Also, please note that you will only get those unexpected values when ScalaCheck actually have found errors in your code with "correct" values. Shrinking is only performed after an error has been found, so if you find and fix the errors in your code (or property), you will not get unexpected values.

ashawley:
Closing as a duplicate of the long-standing issue #129 where shrinking disregards any filters on an explicit Gen.

Issue #338, opened 2017-06-16
Gen.chooseNum(0, Int.MaxValue) sometimes gives -1073741823


shawjef3:
I would expect this to not give any negative values.

shawjef3:
As a work around, I'm using `Gen.choose(0, Int.MaxValue)`.

shawjef3:
Actually, `Gen.choose` isn't a good workaround, because it doesn't test extrema.

shawjef3:
I'm using scalacheck 1.13.5 for scala 2.11.

Philippus:
could you share an example? So far, I cannot reproduce this.

ashawley:
Agree.  I cant 'get the most naive test of this to fail:

```scala
    Prop.forAll(Gen.choose(0, Int.MaxValue)) { x: Int =>
      0 <= x && x <= Int.MaxValue
    }
```

When I intentionally modified the test so it fails, I did see that the shrinking could produce a negative.  

```
> ARG_0: -1
> ARG_0_ORIGINAL: 1967653363
```

That's a known issue with shrinkers, see #129, not with the generators.

pford19:
Curious about the issue tracking and lifecycle process. Why is this issue still in Open status?

rickynils:
@pford19 Lack of time. I want to bring in more maintainers and get better project management for ScalaCheck, but I have unfortunately not time for that either.

pford19:
:-)

SethTisue:
sounds like maybe we should just close this, since there's no clear issue here that isn't #129?

non:
Agreed, I think that issue tracks the problem, and `-disableLegacyShrinking` is a work around for it.

plokhotnyuk:
@non am I understand correctly that disabling of legacy shrinking is not turned on by default in latest versions?

PR #522, opened 2019-08-28
Comprehensive review of Test.Parameters.

### LINKED ISSUE #360: How should Properties.overrideParameters work?

non:
This commit does several things:

  1. Ensures that overrideParameters applies *after* cmdline
  2. Fixes initialSeed to work with Test.Parameters
  3. Adds parameter to disable Shrink-based shrinking
  4. Fixes numerous bugs around cmdline parsing
  5. Adds tests to ensure Test.Parameters is working
  6. Internal clean up and refactoring

This commit includes the work from #463 and subsumes it. I believe it also fixes #360, #289, and #221.

Given how broken parameters are I think this change is a strict improvement. Without making bigger breaking changes I think it will be hard to "rationalize" how parameters work in ScalaCheck, but I think this gets us _much_ closer to having a story that makes sense.
  
Using the `-disableLegacyShrinking` argument to disable shrinking, this PR can partially mitigate #443,  #432, #347, #338, #317, and #129.

### DIFF
--- a/jvm/src/main/scala/org/scalacheck/Platform.scala
+++ b/jvm/src/main/scala/org/scalacheck/Platform.scala
@@ -41,7 +41,7 @@ private[scalacheck] object Platform {
     else {
       import concurrent._
       val tp = java.util.concurrent.Executors.newFixedThreadPool(workers)
-      implicit val ec = ExecutionContext.fromExecutor(tp)
+      implicit val ec: ExecutionContextExecutor = ExecutionContext.fromExecutor(tp)
       try {
         val fs = List.range(0,workers) map (idx => Future {
           params.customClassLoader.map(
--- a/jvm/src/test/scala/org/scalacheck/CogenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/CogenSpecification.scala
@@ -35,13 +35,13 @@ object CogenSpecification extends Properties("Cogen") {
     }
 
     // Two exceptions are equal if they have the same string representation.
-    implicit val exceptionEqual = new Equal[Exception] {
+    implicit val exceptionEqual: Equal[Exception] = new Equal[Exception] {
       override def equal(a1: Exception, a2: Exception): Boolean =
         a1.toString == a2.toString
     }
 
     // Two throwables are equal if they have the same string representation.
-    implicit val throwableEqual = new Equal[Throwable] {
+    implicit val throwableEqual: Equal[Throwable] = new Equal[Throwable] {
       override def equal(a1: Throwable, a2: Throwable): Boolean =
         a1.toString == a2.toString
     }
@@ -57,7 +57,7 @@ object CogenSpecification extends Properties("Cogen") {
 
   // Avoid reimplementing equality for other standard classes.
   trait EqualLowPriority {
-    implicit def universal[A] = new Equal[A] {
+    implicit def universal[A]: Equal[A] = new Equal[A] {
       override def equal(a1: A, a2: A): Boolean = a1 == a2
     }
   }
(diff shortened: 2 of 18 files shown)


charleso:
💯 for a `disableLegacyShrinking` option. That's a great idea, I wish I had thought of that.

That said, I'm just curious if it's intended to be the same/similar to my #444 PR? The only reason I ask is that it _appears_ (but I'm not 100% sure) to disable shrinking for `Arbitrary` as well, which I would argue is actually working fine. I guess the question is - what do you mean by "legacy" shrinking? I'm happy to talk this offline/elsewhere, I don't want to bog this down with another shrinking discussion here if that's unhelpful.

charleso:
@non Thanks for the reply/clarification, I appreciate it.

>  I'm not convinced that authors of Shrink instances have been careful to avoid these pitfalls. 

Oh for sure. That's not to mention anyone who might have locally overridden an Arbitrary implicit (instead of using forAll with `Gen`).

I guess I was taking a somewhat optimistic view that normal usage with the ScalaCheck provided `Arbitrary`/`Shrink` instances (ie `Int`, `String`, `List`), which should (in theory) line up. From my experience most people don't write their own `Shrink` instances anyway.

> although I agree with you that the vast majority of bugs we see are related to Gen not Arbitrary

Cool. I guess I still wasn't 100% sure people didn't think I was crazy/wrong about that point. :)

PR #523, opened 2019-08-29
Test.Parameters.cpy should be private

ashawley:
The previous version, `cp`, was private.  I don't think there's a reason to expose this: 
 `Test.Parameters.cpy` should be enough of a fluent interface that a copy constructor is unnecessary.

### DIFF
--- a/src/main/scala/org/scalacheck/Test.scala
+++ b/src/main/scala/org/scalacheck/Test.scala
@@ -132,7 +132,8 @@ object Test {
       sb.toString
     }
 
-    def cpy(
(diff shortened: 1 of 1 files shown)


non:
Yes! Good catch! Could it be `private[this]` (i.e. even more private)?

non:
Looks like `private[this]` does work (I tested locally), and it was already being used for the `cpy` method in `Gen`. Feel like making that change?

ashawley:
Sure, I will edit to make `private[this]`

PR #526, opened 2019-08-30
Draft release notes for 1.14.1

ashawley:


non:
@ashawley This looks good! One suggestion is in addition to contributors we should mention that you're now a ScalaCheck maintainer.

I also want to write a prose description of the shrinking situation, but I can do that in a follow-up.

SethTisue:
> One suggestion is in addition to contributors we should mention that you're now a ScalaCheck maintainer.

I was going to make a similar suggestion — in addition to the laundry list of technical changes, I think it would be appropriate to describe the change of maintenance status. it might help attract fresh contributors and fresh thinking

ashawley:
I had a similar idea of mentioning the organizational changes.  Hopefully, publishing a release is communication enough that things are active again.

non:
LGTM 👍

non:
I'm fine merging this now, although I thought of two more things:

1.  I think we could consider adding the newest bug fix.
2. I think we should add maintainers to the list of contributors

It might seem weird to suggest (2) but I think it's nice to make the determination of who gets included there (who contributed PRs/commits) as standardized as possible. It's also true that not every maintainer is guaranteed to contribute to every release.

ashawley:
Ok, I've added ourselves per (2).  

I didn't think we needed to add (1), since I assumed it was just follow-up to enabling `viewSeed` by default, which didn't exist in 1.14.0.  However, it's a good suggestion, because if people had been using `viewSeed`, then they could have seen this defect.

non:
These looks great! 👍 

Merge when ready.

ashawley:
Thanks for the suggestions.  Obviously, feel free to add more or further revise anything.

PR #529, opened 2019-08-30
Fix output for bad command-line params

### LINKED ISSUE #527: The output for bad options shows List
```
sbt:scalacheck> jvm/test:runMain org.scalacheck.PropSpecification -h
[info] Done packaging.
[info] Running (fork) org.scalacheck.PropSpecification -h
[info] Incorrect options: List(-h)
[info] Available options:
[info]   -maxDiscardRatio, -r: The maximum ratio between discarded and succeeded tests allowed before ScalaCheck stops testing a property. At least minSuccessfulTests will always be tested, though.
[info]   -verbosity, -v: Verbosity level
[info]   -propFilter, -f: Regular expression to filter properties on
[info]   -workers, -w: Number of threads to execute in parallel for testing
[info]   -maxSize, -x: Maximum data generation size
[info]   -disableLegacyShrinking: Disable legacy shrinking using Shrink instances
[info]   -initialSeed: 
[info]   -minSize, -n: Minimum data generation size
[info]   -minSuccessfulTests, -s: Number of tests that must succeed in order to pass a property
```

ashawley:
This fixes #527 but also switches to `Console.out.println` in a few places, and removes a strange use of `printf` and carriage returns in the `ConsoleReporter`.  I don't entirely grok the latter, but I don't think it serves any purpose.

### DIFF
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -104,7 +104,7 @@ sealed abstract class Prop extends Serializable { self =>
         if (Test.check(params, this).passed) 0
         else 1
       case (_, os) =>
-        println("Incorrect options:\n  " + os.mkString(", "))
+        Console.out.println("Incorrect options:\n  " + os.mkString(", "))
         Test.CmdLineParser.printHelp()
         -1
     }
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -104,7 +104,7 @@ sealed abstract class Prop extends Serializable { self =>
         if (Test.check(params, this).passed) 0
         else 1
       case (_, os) =>
-        println(s"Incorrect options: $os")
+        println("Incorrect options:\n  " + os.mkString(", "))
         Test.CmdLineParser.printHelp()
         -1
     }
--- a/src/main/scala/org/scalacheck/Properties.scala
+++ b/src/main/scala/org/scalacheck/Properties.scala
@@ -73,13 +73,13 @@ class Properties(val name: String) {
         val res = Test.checkProperties(params, this)
         val numFailed = res.count(!_._2.passed)
         if (numFailed > 0) {
-          println(s"Found $numFailed failing properties.")
+          Console.out.println(s"Found $numFailed failing properties.")
           System.exit(1)
         } else {
           System.exit(0)
         }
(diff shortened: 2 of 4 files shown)


ashawley:
Output for bad arguments that are passed:

```
sbt:scalacheck> jvm/test:runMain org.scalacheck.PropSpecification -h -v 2.0
[info] Running (fork) org.scalacheck.PropSpecification -h -v 2.0
[info] Incorrect options:
[info]   -h
[info]   -v
[info]   2.0
[info] Available options:
[info]   -workers, -w: Number of threads to execute in parallel for testing
[info]   -minSize, -n: Minimum data generation size
...
```

non:
This is an improvement but I think there are two things we still need to improve:

1. For the "Incorrect options:" list I think we should either put them all on one line. Most options take an argument as well (e.g. `-v 2`) so displaying each argument on its own line is more difficult to visually parse in my opinion (at least in your example).

2. Similarly, when printing the "Available options:" list we should probably indicate which options take an argument, and broadly what it is, e.g.

```
[info]   -verbosity, -v INT: Verbosity level
[info]   -initialSeed SEED: Use Base-64 seed for all properties
[info]   -maxDiscardRatio, -r FLOAT: The maximum ratio between discarded and succeeded tests allowed before ScalaCheck stops testing a property. At least minSuccessfulTests will always be tested, though.
[info]   -propFilter, -f REGEX: Regular expression to filter properties on
[info]   -disableLegacyShrinking: Disable legacy shrinking using Shrink instances
[info]   -workers, -w INT: Number of threads to execute in parallel for testing
[info]   -maxSize, -x INT: Maximum data generation size
[info]   -minSize, -n INT: Minimum data generation size
[info]   -minSuccessfulTests, -s INT: Number of tests that must succeed in order to pass a property
```

Would you be willing to handle (1) here? I'm willing to handle (2) in a follow up if you want.

ashawley:
I agree about (1).   I've put the options on the same line:

```
sbt:scalacheck> jvm/test:runMain org.scalacheck.PropSpecification -s 4 -v 2.0 -h
Running (fork) org.scalacheck.PropSpecification -s 4 -v 2.0 -h
Incorrect options:
  -v, 2.0, -h
Available options:
  -workers, -w: Number of threads to execute in parallel for testing
  -minSize, -n: Minimum data generation size
...
```

Yeah, (2) is a good improvement, as well.

non:
👍

PR #535, opened 2019-09-14
Set up a repeatable release process.

non:
I had to make a few tweaks to build.sbt to get things reliably building for me
in a sustainable way. Once I got that working, I set up a release.sh script
that I can successfully use to cross-build, cross-test, and cross-publish.

The following commands are useful during the release process:

```
    ./release.sh compile        # build along all axes
    ./release.sh test           # ensure all tests pass
    ./release.sh package        # ensure we can package everythign
    ./release.sh publishSigned  # actually publish a signed release
```

Note that to publish you currently have to (temporarily) modify build.sbt to
change a few variables. You also have to log into Sonatype to close/publish the
release (assuming all its checks pass).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -145,6 +145,8 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
 lazy val js = project.in(file("js"))
   .settings(sharedSettings: _*)
   .settings(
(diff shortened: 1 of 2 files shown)


non:
@ashawley I tagged you on this but I plan to merge it as soon as it's green since these changes correspond to what was actually published as 1.14.1-RC1.

PR #543, opened 2019-09-19
Fix sbt build for examples

ashawley:
According to https://github.com/typelevel/scalacheck/pull/534#issuecomment-532786025, this will be maintained by Scala-Steward going forward.


non:
👍

PR #548, opened 2019-09-20
Auto-set isRelease in release script

ashawley:
Hey Erik,

The release script looks sharp.  The only thing I can think of was making the script handle toggling the `isRelease` value.  There's a benefit to the manual step that you had, but it's possible to automate it.  No hard feelings if you want to keep the manual step.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ scalaVersionSettings
 
 lazy val versionNumber = "1.14.2"
 
-lazy val isRelease = false
+val isRelease = SettingKey[Boolean]("isRelease")
 
 lazy val travisCommit = Option(System.getenv().get("TRAVIS_COMMIT"))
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -25,14 +25,16 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
 
   name := "scalacheck",
 
+  isRelease := false,
```
non: I'm assuming that the command-line overrides this?

release.sh
```diff
@@ -35,23 +33,15 @@ runsbt() {
     sbt "$1"
     RES=$?
     if [ $RES -ne 0 ]; then
-        echo "sbt '$1' failed: $RES"
+        echo "sbt 'set every isRelease := true' '$1' failed: $RES"
```
ashawley: @non Yeah, the shell script overrides it here.

build.sbt
ashawley: Yes, exactly.

non:
👍

PR #549, opened 2019-09-20
Use default Native and JS versions in release script

ashawley:
Updating the version in the script could fall behind, so there is a benefit to defaulting here since Scala Steward probably can't grok the shell script.  

There is a benefit to making it explicit here and for manually checking ScalaJS and Native versions during a release, so happy to just close.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -63,11 +63,11 @@ runsbt "+ clean"
 runsbt "+ jvm/$CMD"
 
 # step 4b: js releases (clean versions)
-SCALAJS_VERSION="0.6.28" runsbt "+ js/$CMD"
+runsbt "+ js/$CMD" # Scala.js 0.6.x
 runsbt "+ js/clean"
 SCALAJS_VERSION="1.0.0-M8" runsbt "+ js/$CMD"
 
 # step 4c: native releases (clean versions)
-SCALANATIVE_VERSION="0.3.9" runsbt "+ native/$CMD"
+runsbt "+ native/$CMD" # Scala Native 0.3.x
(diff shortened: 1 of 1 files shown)


SethTisue:
> There is a benefit to making it explicit here and for manually checking ScalaJS and Native versions during a release

I'm not sure I see the benefit

ashawley:
If you print the shell script on letter paper, turn it sideways, and hold it up the window and let the sunlight...

ashawley:
I wanted to propose this change, but I presume there is a reason Erik went this route.

I've also ever seen people get confused by the empty `SCALAJS_VERSION` logic.  You dabble in these scripts all over the place in the Scala ecosystem, so you have the build matrix integrated to your cerebellum.

SethTisue:
okay. I like this change, but Erik is the one who actually does the releases and understands it best, so let's see what he thinks.

ashawley:
I'd also prefer to follow the DRY principle, but I'm friendly to both options.

non:
@ashawley So, my rationale here (which may be flawed) is that because we're required to hardcode a ScalaJS version in one of the cases (e.g. 1.0.0-M8), it's better to hardcode it in both cases so that we know we have to update the install script whenever any of them change.

I could imagine adding validation in `plugins.sbt` that the environment variable is a known-value. That way it would be impossible for that file to get out of sync with the install script.

(If we had some way of cross-building across plugins that would be ideal but I'm not going to hold my breath.)

ashawley:
Yeah, I agree.  This only provides half of the DRY benefits.

We can use #550, instead.

PR #550, opened 2019-09-20
Update Scala.js 0.6.29 in release script

ashawley:
This is the alternative to #549.


non:
I commented over at #549. I prefer this (possibly with the additional check to validate the environment variable).

ashawley:
That is fine.  I'm not sure how to validate the environment variable.  The code comments in `plugins.sbt` is the best I could come up with.

non:
👍

PR #551, opened 2019-09-20
Validate ScalaJS and Scala Native versions in plugins.sbt.

non:
Since we pass explicit versions in release.sh, it's important to make sure it
doesn't get out of sync with our definitions in plugins.sbt. To that end, this
commit ensures that we get one of two exact versions we're prepared to support.
Otherwise, we'll throw an error during SBT initialization.

This makes it slightly more annoying to test ScalaCheck with a new version of
one of these plugins (you have to modify plugins.sbt as opposed to just passing
a new variable on the command line) but also means we'll never accidentally
release with the wrong version.

I ended up printing before throwing an error because SBT doesn't display the
actual exception (and message) unless you type (l)ast. This way, you'll always
see the message (which should make it pretty obvious what the mistake was).

### DIFF
--- a/project/plugin.sbt
+++ b/project/plugin.sbt
@@ -2,15 +2,29 @@ addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.6.1")
 
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "2.0.0")
 
+def env(name: String): Option[String] =
+  Option(System.getenv(name))
+
+def printAndDie(msg: String): Nothing = {
+  println(msg)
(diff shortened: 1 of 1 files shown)


ashawley:
Yeah, this avoids releasing the wrong version of Native and Scala.js.  The only thing I can think of is that this likely closes the door on Scala Steward being able to maintain these settings.

non:
@ashawley Yeah, it's a shame. OTOH presumably we'll eventually be able to stop cross-publishing for 0.6x and 1.0.x (or their Native equivalents).

PR #552, opened 2019-09-20
Re-enable snapshot releases?

ashawley:
Erik pointed out to me that snapshots can be published to Sonatype.  Any interest in re-enabling this?  It may be useful for 1.15 development.  I think it might have been accidentally disabled during Travis cleanup.

Humorously, it was still occuring for just Scala Native:

https://oss.sonatype.org/content/repositories/snapshots/org/scalacheck/scalacheck_native0.3_2.11/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,9 @@ env:
     - PLATFORM=jvm SBT_PARALLEL=true  WORKERS=4 DEPLOY=false
     - PLATFORM=js  SBT_PARALLEL=true  WORKERS=1 DEPLOY=false
     - PLATFORM=js  SBT_PARALLEL=true  WORKERS=1 DEPLOY=false SCALAJS_VERSION=1.0.0-M8
+    - PLATFORM=jvm SBT_PARALLEL=true  WORKERS=4 DEPLOY=true
+    - PLATFORM=js  SBT_PARALLEL=true  WORKERS=1 DEPLOY=true
+    - PLATFORM=js  SBT_PARALLEL=true  WORKERS=1 DEPLOY=true  SCALAJS_VERSION=1.0.0-M8
 sudo: false
 
 before_cache:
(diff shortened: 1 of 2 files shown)


ashawley:
I think the problem was trying to avoid deploying Java 11, so this needs more work.

ashawley:
I copied the build page from an old Travis build and compared it with the build here:

```diff
@@ -1,14 +1,14 @@
 JDK: oraclejdk8 Scala: 2.10.7
-PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=false
+PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=true
 
 JDK: oraclejdk8 Scala: 2.11.12
-PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=false
+PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=true
 
 JDK: oraclejdk8 Scala: 2.12.10
-PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=false
+PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=true
 
 JDK: oraclejdk8 Scala: 2.13.1
-PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=false
+PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=true
 
 JDK: openjdk11 Scala: 2.12.10
 PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=false
@@ -17,22 +17,22 @@
 PLATFORM=jvm SBT_PARALLEL=true WORKERS=4 DEPLOY=false
 
 JDK: oraclejdk8 Scala: 2.10.7
-PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=false
+PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=true
 
 JDK: oraclejdk8 Scala: 2.11.12
-PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=false
+PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=true
 
 JDK: oraclejdk8 Scala: 2.12.10
-PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=false
+PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=true
 
 JDK: oraclejdk8 Scala: 2.13.1
-PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=false
+PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=true
 
 JDK: oraclejdk8 Scala: 2.12.10
-PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=false SCALAJS_VERSION=1.0.0-M8
+PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=true SCALAJS_VERSION=1.0.0-M8
 
 JDK: oraclejdk8 Scala: 2.13.1
-PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=false SCALAJS_VERSION=1.0.0-M8
+PLATFORM=js SBT_PARALLEL=true WORKERS=1 DEPLOY=true SCALAJS_VERSION=1.0.0-M8
 
 JDK: oraclejdk8 Scala: 2.11.12
 PLATFORM=native SBT_PARALLEL=true WORKERS=1 DEPLOY=true

```

non:
Thanks for setting this up again!

Before we turn it back on, I'd like to figure out who uses these and how much impact they have? I think they do slow down the CI build and also use up space on Sonatype. @SethTisue do you end up using these with the community build? Do we know of anyone who routinely uses these?

SethTisue:
> do you end up using these with the community build? 

nope. (the community build builds everything from source)

ashawley:
They've been uploaded and available for four years, but I don't think many people knew about the snapshots.    The only example of using a snapshot that I ran across was the scalatestplus-scalacheck repo.  Coincidentally, that was for the Native part of their build.

Looking at recent uploads of Native, it takes about 30s to run the travis-deploy script.  Some of that is sbt startup.  Turning it back on here would add 10 more snapshot uploads.  Since Travis runs a several parallel builds, it would seem to increase the build time by a a minute or so.  For comparison, the wall time of the Travis build is approximately 20 minutes.

non:
OK, let's merge this. If the build _does_ get a lot more unstable or slow, we can always revert.

ashawley:
Yes, if it becomes a problem, then we could back it out.

ashawley:
Let's hold off on this for a moment.  I predict that we'd be publishing two Scala.js 0.6 artifacts for Scala 2.13, so we need to deal with #555 and #556.

Issue #555, opened 2019-09-20
Publish 1.14.1 for Scala.js 0.6 and Scala 2.10


ashawley:
While trying to bump ScalaCheck in ScalaTest, their CI caught that Scala.js for 0.6 and Scala 2.10 is missing, see https://github.com/scalatest/scalatest/pull/1706

non:
I went ahead and manually published 1.14.1 for 2.10 with JS 0.6.

We should definitely also merge #556 but it's possible this will unblock the ScalaTest PR.

ashawley:
Thanks, Erik.  It looks good.

https://repo1.maven.org/maven2/org/scalacheck/scalacheck_sjs0.6_2.10/1.14.1/

PR #556, opened 2019-09-20
Fix Scala.js 1.0 and Scala 2.10 in build

ashawley:
Related to #555.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,6 +21,9 @@ lazy val scalaVersionSettings = Seq(
   }
 )
 
+lazy val scalaJSVersion =
+  Option(System.getenv("SCALAJS_VERSION")).getOrElse("0.6.29")
+
 lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
 
   name := "scalacheck",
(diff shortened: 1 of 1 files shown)


ashawley:
Travis was trying to run Scala.js 0.6 with Scala 2.10, but sbt would silently fail to set 2.10.7 for the `js` project:

```
[info] Setting Scala version to 2.10.7 on 2 projects.
[info] Excluded 2 projects, run ++ 2.10.7 -v for more details.
```

This would cause the build to just run with 2.13.

The fix here makes Travis run the correct Scala version.  One thing I hadn't considered is MiMa will fail here until 1.14.1 is published for Scala.js 0.6 and Scala 2.10, which see #555.

non:
I'm going to apply this locally and publish 1.14.1 for 2.10 on JS -- I think that's what the build failure is about.

non:
Steps I took to publish 1.14.1 for 2.10 and ScalaJS 0.6.28:

1. Created `1.14.x` branch
2. Cherry-picked this commit and resolved the conflict
3. Read `release.sh`
4. Locally-edited `build.sbt` and set `isRelease` to `true`
5. Ran `SCALAJS_VERSION="0.6.28" sbt`
6. In SBT ran `++ 2.10.7` and then `js/publishSigned`
7. In Sonatype, closed and released the staging repo

This should resolve the immediate problem in #555 and should also allow MiMA checks for JS on 2.10 to start working.

ashawley:
> I think that's what the build failure is about.

Yes, precisely.  I somehow forgot to mention that above.

ashawley:
> should also allow MiMA checks for JS on 2.10 to start working

Yes, I restarted the build and now we're green here. Thanks, Erik!

Issue #557, opened 2019-09-21
Scala.js tests don't run


ashawley:
There is no output when running tests for Scala.js, and the count of tests is zero.  I can't get a Scala.js to fail, so it doesn't seem they are being run.

```
sbt:scalacheck> js/testOnly org.scalacheck.PropSpecification -- -s 1
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[success] Total time: 3 s, completed Sep 21, 2019 9:33:01 AM
```

<details>
<summary>
Checking out the 1.14.0_sonatype branch it works again.
</summary>

```
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
[info] Fast optimizing js/target/scala-2.12/scalacheck-test-fastopt.js
[info] + Prop.lzy: OK, proved property.
[info] + Prop.Prop.++ False: OK, passed 1 tests.
[info] + Prop.someFailing: OK, passed 1 tests.
[info] + Prop.Prop.&& Exception: OK, passed 1 tests.
[info] + Prop.proved: OK, passed 1 tests.
[info] + Prop.Prop.&& Identity: OK, passed 1 tests.
[info] + Prop.atLeastOne: OK, passed 1 tests.
[info] + Prop.undecided: OK, passed 1 tests.
[info] + Prop.exception: OK, passed 1 tests.
[info] + Prop.all: OK, passed 1 tests.
[info] + Prop.Prop.++ Identity 1: OK, passed 1 tests.
[info] + Prop.Prop.&& Right prio: OK, passed 1 tests.
[info] + Prop.delay: OK, proved property.
[info] + Prop.secure: OK, passed 1 tests.
[info] + Prop.Prop.==> short circuit: OK, passed 1 tests.
[info] + Prop.Prop.++ Exception: OK, passed 1 tests.
[info] + Prop.Prop.==> true: OK, passed 1 tests.
[info] + Prop.Prop.&& Undecided: OK, passed 1 tests.
[info] + Prop.prop.useSeed is deterministic: OK, passed 1 tests.
[info] + Prop.Prop.&& Exception 2: OK, passed 1 tests.
[info] + Prop.Prop.|| Commutativity: OK, passed 1 tests.
[info] + Prop.Prop.|| Exception: OK, passed 1 tests.
[info] + Prop.throws: OK, proved property.
[...]

PR #558, opened 2019-09-21
Partial revert of ff83d14

ashawley:
I'm not familiar with this portion of the code base, so I'm not sure what the consequence of reverting this, but this change causes Scala.js tests to run again.


non:
Is there an issue related to JS tests not running. I do see how this change could cause different behavior, although I'm not entirely sure why. I'm comfortable reverting this as long as we have a way of seeing if this fixes a problem (and ideally a unit test).

non:
Aha! I see that this was reported in #557. Sorry for the confusion! I'm fine merging this if it fixes the problem. It's also likely that we might want to cut a 1.14.2 release, since this issue could affect others too.

ashawley:
Yes, this is related to #557.  My fault for failing to mention that.

non:
👍

PR #560, opened 2019-09-21
Release notes for 1.14.2

ashawley:
This will likely be the only change in the release.

This is also an opportunity to fix up whitespace/wordwrap in the release notes.

Contributor list pulled with: `git shortlog -ns --no-merges 1.14.1..HEAD`

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,5 +1,43 @@
 # ScalaCheck CHANGELOG
 
+## 1.14.2 (2019-09-25)
+
+* Binary compatible with 1.14.1 version of ScalaCheck.
+
+### Changed
+
+* No user-visible changes.
+
+### Fixed
+
+* Tests are not being run at all on Scala.js
+  [#557](https://github.com/typelevel/scalacheck/issues/557)
+
+### Added
+
+* No added features.
+
+### Administrivia
+
+* Further improvements to ScalaCheck release script
+
+* Update build to sbt 1.3.2
+
+* Update MiMa plugin to 0.6.1
+
+* Update sbt-pgp to 2.0.0
+
+* Update Scala.js example project
+
+### Contributors
+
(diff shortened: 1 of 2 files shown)


SethTisue:
maybe this was laid out already and I missed it, but:

why bother releasing if there are no user-visible changes?

ashawley:
It's a bug fix release.

SethTisue:
Does #557 affect users, and not just CI?

ashawley:
Yes, #557 is the bug.  Yes, it happens locally for me, not just CI builds.

ashawley:
In case you missed it, Erik mentioned a 1.14.2 release in [a comment in #558](https://github.com/typelevel/scalacheck/pull/558#issuecomment-533824460).

SethTisue:
> Yes, it happens locally for me, not just CI builds

It happens locally for you during development work on ScalaCheck itself, only? Or it also happens when Scala.js users use 1.14.1 in their own projects?

ashawley:
Yes, it happens downstream when running Scala.js tests with 1.14.1.

SethTisue:
okay, clear now. some more direct wording such as "Doesn't work on Scala.js, tests don't run" might be clearer than "Scala.js tests are not being run"

ashawley:
Ok, that's a valid criticism.  I suppose "doesn't work" is right, but I tried to be more specific.  How about "Tests are not being run _**at all**_ on Scala.js"?  Better?

SethTisue:
perhaps it's simply that as a maintainer/contributor, I have CI on the brain. perhaps actual users won't have the same understanding problem I did

in any case: as you see fit, now that we understand each other.

non:
@ashawley Want to mark this as released and merge it now?

ashawley:
Sure thing.  Marked as released and merging.

PR #563, opened 2019-09-26
Bump version in build post-release

ashawley:
- Bump to 1.14.3
- Update MiMa to 1.14.2
- Update examples. to 1.14.2
- Give more recent snapshot example

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ val scalaMajorVersion = SettingKey[Int]("scalaMajorVersion")
 
 scalaVersionSettings
 
-lazy val versionNumber = "1.14.2"
+lazy val versionNumber = "1.14.3"
 
 def env(name: String): Option[String] =
   Option(System.getenv(name))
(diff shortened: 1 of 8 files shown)


non:
@ashawley do you want to target the `1.14.x` branch with this? I'd like to bump master to `1.15.0` so we can start doing things like e.g. removing 2.10.

ashawley:
Sure, I'll work on that.

ashawley:
@non Changing the base brought in a bunch of more changes. I've merged last master in to 1.14.x and pushed up a new 1.14.x, since that's what you had done previously.  Rebasing this onto the new 1.14.x has the single changes in this PR.  Hopefully, that's what you were looking for here.

non:
That sounds fine. I had imagined possibly waiting to backport the other changes from master to 1.14.x until we need them, but this works too.

👍

ashawley:
Do you want me to do a version of this for 1.15.0?  Or were you already working on it?

PR #564, opened 2019-09-26
Clean up the build

ashawley:
- Drop explicit memory settings for JVM
- Only fork JVM in sbt with Scala 2.13
- Switch to Xenial in Travis
- Switch to OpenJDK for Java 8 in Travis
- Update `pomExtra` in sbt to new syntax
- Remove errant `scalacOptions` in `plugins.sbt`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -80,8 +80,6 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
 
   resolvers += "sonatype" at "https://oss.sonatype.org/content/repositories/releases",
 
-  javacOptions += "-Xmx1024M",
-
   // 2.10 - 2.13
   scalacOptions ++= {
     def mk(r: Range)(strs: String*): Int => Seq[String] =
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -171,7 +171,11 @@ lazy val js = project.in(file("js"))
 lazy val jvm = project.in(file("jvm"))
   .settings(sharedSettings: _*)
   .settings(
-    fork in Test := true,
+    fork in Test := {
+      // Serialization issue in 2.13
+      scalaMajorVersion.value == 13 // ==> true
+      // else ==> false
+    },
```
non: Is there a reason we don't want to fork unconditionally?

build.sbt
```diff
@@ -141,18 +139,20 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
...
+    )
+  ),
+  developers := List(
+    Developer(
+      id    = "rickynils",
+      name  = "Rickard Nilsson",
+      email = "rickynils@gmail.com",
+      url   = url("https://github.com/rickynils")
+    )
+  )
```
non: We should probably add ourselves to the developers list at this point.

non:
@ashawley Thanks! This looks good, although I had a couple of questions/comments.

build.sbt
ashawley: I'm not sure.  While working on fixing the 2.13 serialization issue, this was changed to forking unconditionally for all versions.  I assumed it was a change made broadly in the narrow context of needing to publish 2.13.  I'm just trying to revert back.

It would likely be useful to get coverage for both cases, but that may be more growth (a doubling!) to the CI build than we'd want to entertain.

build.sbt
ashawley: That's fine.  I'd rather add ourselves in a follow-up, though.

non: 👍

build.sbt
non: OK, sure, that works.

non:
👍

PR #565, opened 2019-09-26
Compile in dotty 0.19.0-RC1

ashawley:
Previously #562.


ashawley:
It seems the dotty doc tool fails on the user guide here:
```
-- Error: doc/UserGuide.md:1:6 -------------------------------------------------
1 |# ScalaCheck User Guide
  |      ^
  |      unexpected token 'token  57', expected: 'token  77'
```

Issue #568, opened 2019-09-26
Gen.posNum fails to produce a value, listOfN and other combinators don't handle this


zakpatterson:
On 1.14.2 and 1.14.1, I get the following:
```scala
scala> (1 to 10000).toList.map(_ => Gen.posNum[Int].sample).flatten.length
res99: Int = 9904

scala> (1 to 10000).toList.map(_ => Gen.posNum[Int].sample).flatten.length
res100: Int = 9887
```

This issue is absent on 1.14.0:
```scala
scala> (1 to 10000).toList.map(_ => Gen.posNum[Int].sample).flatten.length
res1: Int = 10000

scala> (1 to 10000).toList.map(_ => Gen.posNum[Int].sample).flatten.length
res2: Int = 10000
```

This might be related to #567 
 
[1.14.1 failing build](https://travis-ci.org/clovellytech/http4s-modules/jobs/586623814#L1449)
[1.14.2 failing build](https://travis-ci.com/clovellytech/http4s-modules/jobs/239120610#L1376)
[My project's version pr](https://github.com/clovellytech/http4s-modules/pull/147)
[My use case](https://github.com/clovellytech/http4s-modules/blob/master/modules/store/src/test/scala/h4sm/store/infrastructure/repository/sql/arbitraries.scala#L12)

non:
@zakpatterson This is an intentional result of a change that enables floating point numbers between 0 and 1 to be generated. We could do something more sophisticated to change this behavior for integral types, but for fractional types it's very unlikely we'll be able to avoid throwing away some values.

zakpatterson:
@non interesting. Thanks for the explanation. It seems though other parts of scalacheck are not checking for this behavior, for example Gen.listOfN will just return None forever given a large N:

```scala
scala> Gen.listOfN(10000, Gen.posNum[Int]).sample
res2: Option[List[Int]] = None
```

Continuing the conversation for a moment here:
https://gitter.im/scalacheck/Lobby?at=5d8d1a69fb131014e7f0d74a

non:
@zakpatterson Yeah, I think `listOfN` needs to be able to handle occasional failures to generate without failings its own list. I think there are two issues here:

 - Feature request: make `posNum` always succeed on integral types
 - Bug fix: make `listOfN` (and other combinators) more resilient to failures

zakpatterson:
Ok, I'll take a shot at a PR for at least the first.

zakpatterson:
[...]

PR #569, opened 2019-09-26
Force posNum and negNum generators to always return values

zakpatterson:
Fixes (part of) #568.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,5 +1,9 @@
 # ScalaCheck CHANGELOG
 
+### Changed
+
+* Force posNum and negNum to retry if they produce an illegal value and otherwise would produce None, since 1.14.1.
+
 ## 1.14.2 (2019-09-25)
 
 * Binary compatible with 1.14.1 version of ScalaCheck.
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,8 +1,9 @@
 # ScalaCheck CHANGELOG
 
-### Changed
+### Fixed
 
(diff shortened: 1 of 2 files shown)


non:
This looks good. Would you be willing to add one other change? I think unless our `Numeric[T]` is actually a `Fractional[T]` we can do something even simpler, i.e.

```scala
def posNum[T](implicit num: Numeric[T], c: Choose[T]): Gen[T] = {
  import num._
  num match {
    case _: Fractional[_] =>
      sized(n => c.choose(zero, max(fromInt(n), one)).suchThat(_ != zero))
    case _ =>
      sized(n => c.choose(one, max(fromInt(n), one))
  }
}
```

The reason this might be better is that if `n` is one, the fractional types will still be able to generate plenty of non-zero values, but the integral types will be generating zero 50% of the time (resulting in more retries).

zakpatterson:
Yes, that makes sense. I'm playing with it now and I'll push tonight.

zakpatterson:
While I'm staring at this, is there a reason that negNum is not just 

```scala
  def negNum[T](implicit num: Numeric[T], c: Choose[T]): Gen[T] = posNum[T].map(num.negate(_))
```

non:
@zakpatterson Using negation on negative numbers can be dangerous (since `-Int.MinValue == Int.MinValue`) but I think your change makes sense.

This looks good. Thanks! 👍

zakpatterson:
Thanks for the review @non, I learned something.

non:
I think the build failure was transient, I'm retrying now.

PR #571, opened 2019-09-27
Fix indentation of codegen

ashawley:
Follow-up to #517.

Even after the change here, the same generated code in `jvm/target/scala-2.13/src_managed` was unchanged.


non:
👍

PR #590, opened 2019-12-12
Add Cogen.domainOf

### LINKED ISSUE #589: Add a function similar to Gen.resultOf to Cogen
Currently to build a `Cogen` for some type (`A`) that is isomorphic to another type (`B`) for which a `Cogen` exists, we need to summon the `Cogen[B]`, and `contramap` it. This leads to noisy code like this:

```scala
implicit val InstantCogen: Cogen[Instant] =
  Cogen[(Long, Int)].contramap(i => (i.getEpochSecond, i.getNano))
```

It should be easy to add `Cogen.resultOf` that would derive a `Cogen[A]` given a function `A => B` (and the implicit `Cogen[B]` in scope), similarly to what `Gen.resultOf` does. We would then be able to write:

```scala
implicit val InstantCogen: Cogen[Instant] =
  Cogen.resultOf(i => (i.getEpochSecond, i.getNano))
```

nigredo-tori:
Closes #589

I'm not sure about the name. `resultOf` doesn't fit (we're creating `Cogen` for the domain of the function). I'm using `domainOf` for now, but I'm open to suggestions.

I didn't add any tests since this is just sugar over `contramap`.

### DIFF
--- a/src/main/scala/org/scalacheck/Cogen.scala
+++ b/src/main/scala/org/scalacheck/Cogen.scala
@@ -167,6 +167,8 @@ object Cogen extends CogenArities with CogenLowPriority with CogenVersionSpecifi
     while (i < as.length) { s = A.perturb(s, as(i)); i += 1 }
     s.next
   }
+
+  def domainOf[A, B](f: A => B)(implicit B: Cogen[B]): Cogen[A] = B.contramap(f)
 }
 
 trait CogenLowPriority {


non:
I think the name `by` might be better (consistent with `Ordering.by` which is the same syntactic sugar over contramap) but I'm in favor of merging this now and worrying about it later.

What do you think @ashawley?

ashawley:
I'm fine with `domainOf`.  It seems better than `by` even though that was a natural choice for `Ordering`.  It's too bad we don't have a good Scala style guide for naming factory methods.  My sense is that Java has been converging on using `of`, see this Stephen Colebourne blog post on [Factory Names](https://blog.joda.org/2011/09/factory-names.html).

non:
I had to revert the merge of this PR, but I cherry-picked the commit and pushed it to _master_ instead.

Thanks!

nigredo-tori:
@non, sorry to resurrect this... I can't seem to find `domainOf` in master - or anywhere, really. I only see commits mentioning it in the 1.14.x branch (this PR and #655).

SethTisue:
@nigredo-tori seems like an oversight. would you mind submitting a new PR?

PR #596, opened 2019-12-13
Bump version of artifact for MiMa

ashawley:


ashawley:
MiMa still fails on Scala.js, I'll fix in a follow-up.

PR #616, opened 2020-02-01
Add Gen.fix to make recursive generators cleaner

johnynek:
same idea as https://github.com/typelevel/cats/pull/3208 but particular to Gen.

Gen and Parsers are what motivated the 3208 PR.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -420,6 +420,28 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
     }
   }
 
+  property("fix is like lzy") = forAll { (seeds: List[Seed]) =>
+    lazy val lzyGen: Gen[List[Int]] = {
+      Gen.choose(0, 10).flatMap { idx =>
+        if (idx < 5) lzyGen.map(idx :: _)
+        else Gen.const(idx :: Nil)
+      }
+    }
+
+    val fixGen =
+      Gen.fix[List[Int]] { recurse =>
+        Gen.choose(0, 10).flatMap { idx =>
+          if (idx < 5) recurse.map(idx :: _)
+          else Gen.const(idx :: Nil)
+        }
+      }
+
+    val params = Gen.Parameters.default
+    seeds.forall { seed =>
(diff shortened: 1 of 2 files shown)


ashawley:
What's the example do?  It's building lists that terminate with numbers higher than 5?

```scala
val fixGen =
  Gen.fix[List[Int]] { recurse =>
    Gen.choose(0, 10).flatMap { idx =>
      if (idx < 5) recurse.map(idx :: _)
      else Gen.const(idx :: Nil)
    }
  }
```

johnynek:
Yes. I didn’t want to make a longer example such as a tree or JSON or something in the comment and I wanted to test the same thing that I put in the comment.

It’s probably not a great example because in this case there are ways to do this without recursion.

ashawley:
Yeah, I was also wondering if there was a better example to use.  I do like this example for one reason:  Writing with existing combinators is clunky, because I tried:

```scala
val fixGen = {
  Gen.nonEmptyListOf(Gen.choose(0, 9)).map { ns =>
    val idx = ns.indexWhere(_ >= 5)
    ns.take(if (idx == -1) ns.size - 1 else idx + 1)
  }
}
```

Maybe there is a built-in generator that could be rewritten using this syntax as an example?

ashawley:
This construct seems pretty useful.  Should we elevate it by giving it a better name for visibility sake? Could we rename to something like `from`, `of`, `until` or `continually`?:

```scala
val listGen = Gen.of { recurse: Gen[List[Int]] =>
  ...
}
```

johnynek:
Fix is the name for this traditionally: fix(f) = f(fix(f)).

I think the name should explain something about recursion. If you aren’t doing recursion you don’t need this.

What about `def recursive`?

ashawley:
I agree.  I guess I prefer a verb or an adverb, so even just `recur`.  We can stew on it for a few days.  This could be part of 1.15.0 which should be ready soon.

PS. I'm familiar with fixed point since I used to Scheme occasionally.  I won't suggest `Gen.y`, though.

ashawley:
After some time, I haven't had any strong feelings about a better name, so should we go with `Gen.recursive`?  It works for me.

larsrh:
fast-check calls it `letrec`

ashawley:
That would seem to give evidence for naming it `recursive` then.  I'm going to merge and follow-up with a commit that renames it.

PR #620, opened 2020-02-05
Upgrade Scala.js to 1.0.0

xerial:
Finally Scala.js 1.0.0 has been released https://github.com/scala-js/scala-js/releases/tag/v1.0.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,10 @@ env:
   matrix:
     - PLATFORM=jvm SBT_PARALLEL=true  WORKERS=4 DEPLOY=false
     - PLATFORM=js  SBT_PARALLEL=true  WORKERS=1 DEPLOY=false
-    - PLATFORM=js  SBT_PARALLEL=true  WORKERS=1 DEPLOY=false SCALAJS_VERSION=1.0.0-RC2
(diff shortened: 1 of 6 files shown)


ashawley:
Would you be willing to update the Scala.js example to 1.0 while you're at it?

xerial:
Sure. Let me try :)

xerial:
@ashawley Done. The next target version is 1.14.3 or 1.15.0?

ashawley:
Thanks!

The current version is 1.14.3, the next version will be 1.15.0.  

We can look in to backporting to 1.14.x and pushing a tag like we did most recently in #592, for example.

ashawley:
LGTM!  The mima check fails because 1.14.3 artifacts don't exist, but we can resolve that as I mentioned previously.

larsrh:
It would be great if you could release 1.14.3 with sjs 1.0 support. That way, it wouldn't trigger a wave of version bumps across the ecosystem.

ashawley:
Yes, that's what is being worked on in #623.

ashawley:
1.14.3 artifacts for Scala.js have been released.

xerial:
@ashawley Awesome. Thank you so much 😄

PR #627, opened 2020-02-13
Adds helper to create Shrink using LazyList

### LINKED ISSUE #626: Deprecation warnings because of Stream
I am currently porting code to scala 2.13 and am not able to get rid of deprecation warnings concerning Stream (replaced by LazyList) because of it's use in the interface of Shrink.

Is there a way to do so? A second apply method in Shrink would be handy:
  def apply[T](s: T => LazyList[T]): Shrink[T] = ???

Or am I missing something?

froth:
This allows to use Shrink from scala 2.13 without getting deprecation
warnings on scala.collection.immutable.Stream

fixes #626

### DIFF
--- a/src/main/scala-2.13+/org/scalacheck/ScalaVersionSpecific.scala
+++ b/src/main/scala-2.13+/org/scalacheck/ScalaVersionSpecific.scala
@@ -38,3 +38,8 @@ private[scalacheck] trait CogenVersionSpecific {
   implicit def cogenLazyList[A: Cogen]: Cogen[LazyList[A]] =
     Cogen.it(_.iterator)
 }
+
+private[scalacheck] trait ShrinkVersionSpecific {
+  /** Forward to Shrink instance factory */
+  def withLazyList[T](s: T => LazyList[T]): Shrink[T] = Shrink(s.andThen(_.toStream))
+}
--- a/src/main/scala-2.13-/org/scalacheck/ScalaVersionSpecific.scala
+++ b/src/main/scala-2.13-/org/scalacheck/ScalaVersionSpecific.scala
@@ -37,3 +37,5 @@ private[scalacheck] trait CogenVersionSpecific
 private[scalacheck] trait GenSpecificationVersionSpecific {
   def infiniteLazyList[T](g: => Gen[T]): Gen[Stream[T]] = Gen.infiniteStream(g)
 }
+
+private[scalacheck] trait ShrinkVersionSpecific
--- a/src/main/scala/org/scalacheck/Shrink.scala
+++ b/src/main/scala/org/scalacheck/Shrink.scala
@@ -24,7 +24,7 @@ trait ShrinkLowPriority {
   implicit def shrinkAny[T]: Shrink[T] = Shrink(_ => Stream.empty)
 }
 
-object Shrink extends ShrinkLowPriority {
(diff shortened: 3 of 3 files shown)


mpilquist:
Awesome! I'd personally prefer this as a method directly on `Shrink` companion (via the normal pattern of a `ShrinkPlatform` trait that's mixed in to the companion and defined as an empty trait in 2.12 and prior and a non-empty trait on 2.13.

ashawley:
Yes, that would be the preference.  Though, I don't think it's possible to have both definitions of `apply`  on 2.13 with different override arguments for the function type (`T => Stream[T]` vs. `T => LazyList[T]`).

froth:
> Though, I don't think it's possible to have both definitions of `apply` on 2.13 with different override arguments for the function type (`T => Stream[T]` vs. `T => LazyList[T]`).

You are right. Although to be honest, I don't get why the compiler can not differentiate between Function1[T, Stream[T]] and Function1[T, LazyList[T]].

I followed your suggestions (and am sorry for not looking at the codebase thoroughly before). And I renamed the function form apply to fromLazyList. Again naming is hard, not exactly accurate. Maybe fromLazyListFunction but that seems clunky. Ideas welcome.

ashawley:
I believe it's because Function1 is implemented with type parameters, as well.  There may be a way to overload methods having type parameters with a dummy implicit, but that seems unsavory.

ashawley:
I suggested it, but maybe `fromLazyList` is a poor name.  Perhaps, `withLazyList`?

ashawley:
If we have a name for the Shrink factory method, instead of using `apply`, the interface to Shrink has divered.  Should we deprecate `apply` and add `withStream` for symmetry sake?

mpilquist:
Indeed, `DummyImplicit` can de-conflict the erased overloads if we want it to be called `apply`:

```
def apply[T](s: T => LazyList[T])(implicit d: DummyImplicit): Shrink[T
```

froth:
Actually DummyImplicit does not do the trick on it's own, if I add the apply method with the DummyImplicit I get a lot of errors like the following:
```
[error] /home/froth/projects/scalacheck/src/main/scala/org/scalacheck/Shrink.scala:24:49: missing parameter type                                                                                      
[error]   implicit def shrinkAny[T]: Shrink[T] = Shrink(_ => Stream.empty)  
```
I therefore opted for renaming the factory method to withLazyList. 

> If we have a name for the Shrink factory method, instead of using `apply`, the interface to Shrink has diverged. Should we deprecate `apply` and add `withStream` for symmetry sake?

I don't have an opinion on that, as I do not know what the further plans for LazyList and scalacheck 2.0 are. I am happy to do whatever you decide ;)

ashawley:
Thanks for bearing with us and being amenable to our whims.  I predict the change is pretty benign, since it's just a factory method for Shrink.   It shouldn't paint us in to a corner, and more likely it should be something we can use going forward.

froth:
No worries :) I learned something in the process, which is even dearer to me, then getting rid of those annoying warnings.

ashawley:
I wish there was a way to easily dogfood this new factory in the library. Alternatively, we could add tests, but there aren't many existing example tests for Shrink nor any version-specific tests.

ashawley:
Thanks again for working on this.  We'll wait and see if there are other comments.

SethTisue:
thanks @froth!

PR #634, opened 2020-03-01
Make printing of Commands failures more readable

jonaskoelker:
Example output:

```
[info] > Labels of failing property:
[info] Initial State:
[info]   0
[info] Sequential Commands:
[info]    1. TakeNumber => 1
[info]    2. TakeNumber => 2
[info]    3. TakeNumber => 3
[info] Parallel Commands (starting in state = 3)
[info]    1. TakeNumber => 4
[info]    2. TakeNumber => 5
[info]    3. TakeNumber => 7
[info]
[info]    1. TakeNumber => 6
[info]    2. TakeNumber => 7
[info] ARG_0: ...
[info] ARG_0_ORIGINAL: ...
```

(Here "..." is an abberivation.)

### DIFF
--- a/src/main/scala/org/scalacheck/commands/Commands.scala
+++ b/src/main/scala/org/scalacheck/commands/Commands.scala
@@ -329,26 +329,34 @@ trait Commands {
   }
 
   /** Formats a list of commands with corresponding results */
-  private def prettyCmdsRes(rs: List[(Command,Try[String])]) = {
-    val cs = rs.map {
-      case (c, Success("()")) => c.toString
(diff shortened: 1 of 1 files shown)


SethTisue:
I'm wondering if we should just merge this, even though nobody has stepped forward with a close review. The changes only affect printing, so that doesn't seem high risk.

@jonaskoelker is it worth considering adding test coverage for this...?

ashawley:
Yes, I think this is a low risk change, as well.  Tests would be a good idea, but adding tests for outputs is difficult the way the Commands API is currently written.

PR #635, opened 2020-03-01
Print shrunk values first, then the _ORIGINAL values

jonaskoelker:
This makes `forAll { (x: Int, y: Int) => false }` fail as follows:

[info] > ARG_0: 0
[info] > ARG_1: 0
[info] > ARG_0_ORIGINAL: -1737265162
[info] > ARG_1_ORIGINAL: -1

instead of

[info] > ARG_0: 0
[info] > ARG_0_ORIGINAL: -1737265162
[info] > ARG_1: 0
[info] > ARG_1_ORIGINAL: -1

I prefer the former error message: when a property fails, I typically
want to see a set of inputs that made it fail.  The new layout shows
me first the shrunk set of inputs (with nothing else interleaved),
then the original set of inputs (with nothing else interleaved).

### DIFF
--- a/src/main/scala/org/scalacheck/util/Pretty.scala
+++ b/src/main/scala/org/scalacheck/util/Pretty.scala
@@ -130,13 +130,15 @@ object Pretty {
 
   def prettyArgs(args: Seq[Arg[Any]]): Pretty = Pretty { prms =>
     if(args.isEmpty) "" else {
-      for((a,i) <- args.zipWithIndex) yield {
+      val labeledValues = for((a,i) <- args.zipWithIndex) yield {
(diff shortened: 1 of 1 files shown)


non:
Looks good to me. What do you think @ashawley?

The only downside I can see would be if some tools out there are trying to parse ScalaCheck's output and will get confused. I can't think of anything that does this offhand -- is there something I'm missing?

ashawley:
This makes sense to me.

I doubt there are any tools out there parsing ScalaCheck's failure output, besides the pattern recognition part of a user's brains that is expecting the current format. ;-)

I suppose we'd just need to update documentation and the ScalaCheck book.  The chance of someone noticing the difference are probably small, though.

PR #644, opened 2020-03-11
Use implicit in Shrink.suchThat example

ashawley:
Follow-up to #484.

### DIFF
--- a/doc/UserGuide.md
+++ b/doc/UserGuide.md
@@ -843,12 +843,12 @@ with the condition to be maintained:
 ```scala
 /** Generate lists of an even length */
 val genEvenList: Gen[List[Int]] = Gen.sized { size =>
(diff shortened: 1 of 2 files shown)


ashawley:
I don't think using `implicitly` is possible.  That's probably the reason there aren't a lot of examples for Shrink that use it.  There's no way around that.

koterpillar:
Yes, I wondered that too - perhaps it will be better to make an example with a custom type to avoid possible conflicts?

ashawley:
Yes, while the use of implicits for shrink instances in `forAll` is a convenience, the scoping collisions is a disadvantage.

ashawley:
Yup, Hedgehog has coupled shrink with the generator in their GenT type:

https://github.com/hedgehogqa/scala-hedgehog/blob/28a8a9d/core/shared/src/main/scala/hedgehog/core/GenT.scala

PR #651, opened 2020-05-05
Fix a bug with initial seed support.

non:
Prevously, the initial seed was used for every single property
evaluation, meaning that even if a property is evaluated 100 times the
same values will be used each time. This is not what users expect, and
makes the library harder to use.

(In particular, the MUnit library hit this bug using Test.check.)

After this commit, the initial seed will be used first, but then a
deterministic set of seeds will be used later. This sequence will be
the same for every run as long as the same number of workers are used.
If the number of workers is changed, the initial seed will be the
same, but the exact sequence of seeds used may differ somewhat.

### DIFF
--- a/src/main/scala/org/scalacheck/Test.scala
+++ b/src/main/scala/org/scalacheck/Test.scala
@@ -416,7 +416,7 @@ object Test {
       def isExhausted = d > params.minSuccessfulTests * params.maxDiscardRatio
 
       var seed = {
-        val seed0 = params.initialSeed.getOrElse(rng.Seed.random)
+        val seed0 = params.initialSeed.getOrElse(rng.Seed.random())
         if (workerIdx == 0) seed0 else seed0.reseed(workerIdx.toLong)
         //seed0.reseed(workerIdx.toLong)
       }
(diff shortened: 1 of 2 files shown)


non:
To see this bug in action you can extract the test and run it against master, or use the `-initialSeed` parameter in a ScalaCheck project by adding the following to its `build.sbt` file:

```scala
testOptions in Test +=
  Tests.Argument(TestFrameworks.ScalaCheck, "-initialSeed", "eBdqNuA2XAob9GhQcw0UZVjKQwkrg2cdno3oLj5PeXI=")
```

src/main/scala/org/scalacheck/Test.scala
```diff
@@ -400,37 +401,50 @@ object Test {
...
       var res: Result = null
       var fm = FreqMap.empty[Set[Any]]
 
-      def isExhausted = d > minSuccessfulTests * maxDiscardRatio
+      def isExhausted = d > params.minSuccessfulTests * params.maxDiscardRatio
+
+      var seed = {
+        val seed0 = params.initialSeed.getOrElse(rng.Seed.random())
+        if (workerIdx == 0) seed0 else seed0.reseed(workerIdx.toLong)
+        //seed0.reseed(workerIdx.toLong)
```
ashawley, line 421: Straggling code left as a comment?

ashawley:
Glad to have you back!

I'm not familiar with this portion of the library.  I've noticed this issue, but didn't think it could be easily fixed.  Appreciate that you understand it.

src/main/scala/org/scalacheck/Test.scala
non: Good catch! I'll remove it now.

non:
Thanks, my availability is somewhat limited but I'm starting to be able to do a bit of work in my free time.

PR #654, opened 2020-05-06
Update dotty to 0.24.0-RC1

ashawley:


ashawley:
Compiler error:

```
[error] -- [E049] Reference Error: src/main/scala/org/scalacheck/ScalaCheckFramework.scala:108:8 
[error] 108 |        taskDef.selectors.foreach {
[error]     |        ^^^^^^^
[error]     |Reference to taskDef is ambiguous,
[error]     |it is both defined in method checkPropTask
[error]     |and inherited subsequently in anonymous class ScalaCheckRunner.this.BaseTask {...}
[error] -- [E049] Reference Error: src/main/scala/org/scalacheck/ScalaCheckFramework.scala:146:35 
[error] 146 |          val fullyQualifiedName = taskDef.fullyQualifiedName
[error]     |                                   ^^^^^^^
[error]     |Reference to taskDef is ambiguous,
[error]     |it is both defined in method checkPropTask
[error]     |and inherited subsequently in anonymous class ScalaCheckRunner.this.BaseTask {...}
[error] -- [E049] Reference Error: src/main/scala/org/scalacheck/ScalaCheckFramework.scala:148:28 
[error] 148 |          val fingerprint = taskDef.fingerprint
[error]     |                            ^^^^^^^
[error]     |Reference to taskDef is ambiguous,
[error]     |it is both defined in method checkPropTask
[error]     |and inherited subsequently in anonymous class ScalaCheckRunner.this.BaseTask {...}
[error] -- [E049] Reference Error: src/main/scala/org/scalacheck/ScalaCheckFramework.scala:169:34 
[error] 169 |        val n = if (name.isEmpty) taskDef.fullyQualifiedName else name
[error]     |                                  ^^^^^^^
[error]     |Reference to taskDef is ambiguous,
[error]     |it is both defined in method checkPropTask
[error]     |and inherited subsequently in anonymous class ScalaCheckRunner.this.BaseTask {...}
```

src/main/scala/org/scalacheck/ScalaCheckFramework.scala
```diff
@@ -99,13 +99,13 @@ private abstract class ScalaCheckRunner extends Runner {
       }
   }
 
-  def checkPropTask(taskDef: TaskDef, single: Boolean): BaseTask = new BaseTask(taskDef) {
+  def checkPropTask(taskDef: TaskDef, single: Boolean): BaseTask = new BaseTask(taskDef) { self =>
```
ashawley, line 102: Adding `self =>` seems to fix the compiler error in Dotty 0.24.

Now, I realize the other option is renaming parameter. That would produce signature compatibility issues, but fortunately, this class is `private`.

PR #655, opened 2020-05-07
Revert "Add Cogen.domainOf"

non:
Reverts typelevel/scalacheck#590

This PR was apparently targeting the 1.14.x branch, but I think we should merge it into master instead.

@ashawley What do you think? Did you want this targeting 1.14.x?


ashawley:
Yes, I don't think it should be merged to 1.14.x

PR #656, opened 2020-05-09
Add various statistical distributions to Gen.

non:
This commit adds support for several statistical distributions:

 - Gaussian (normal)
 - Exponential
 - Geometric
 - Poisson
 - Binomial

These distributions are likely to be useful for writing generators, as
well as for fine-grained control of numerical ranges.

Currently, authors using Gen.choose have to decide whether to forbid
testing with large Double values (by using a small range), or to allow
large values (in which case their test will be dominated by large
values). One solution is to use distributions where large values are
possible but rare, such as an exponential distribution.

Similarly, authors can use a geometric distribution to generate the
size of a collection which is usually small, but may occasionally be
large.


ashawley:
Very cool! 👍

PR #662, opened 2020-05-21
Run the StatsSpecification on all targets

ashawley:
Move out of the `jvm` directory, but also rename the filename to plural `Stats...`.


ashawley:
I assume this wasn't intentional by @non, but we'll see if the test succeeds on the Scala.js and native builds.

non:
👍

Issue #540, opened 2019-09-18
next steps: 1.15.0 versus 1.14.2


non:
First, some background:

 - The 1.14.x series will maintain source- and binary-compatibility with 1.14.0.
 - The 1.15.x series will preserve binary-compatibility with 1.14.0 but is expected to **break** source-compatibility (e.g. removing the `implicit` keyword from some definitions)
 - We are planning to drop support for Scala 2.10 in 1.15, but we kept this in for the 1.14.1 release.

Now that 1.14.1 is out, @ashawley and I were discussing next steps. My proposal is that we start merging PRs intended for 1.15.x to `master`. If we don't encounter major bugs in 1.14.1 I'd prefer not to have to do a 1.14.2 release and go right to 1.15.0. If we _do_ need 1.14.2, we can create a `1.14.x` branch off of the `1.14.1` tag at that time and go from there.

How does that sound? Any objections? Am I missing anything?

cc @SethTisue @ashawley et al.

smarter:
i suggest creating a 1.14.x branch just for the record even if nothing more ends up pushed to it.

ashawley:
Yes, a 1.14.x branch would likely be useful.  It's what's been done in the past.

ashawley:
FYI: In light of #557, we will have to consider a 1.14.2 release.

non:
1.14.2 is released -- once we've attached the release notes I'll post about it publicly.

ashawley:
Thanks, Erik, for doing the release.  

I expect in the age of Scala Steward we probably get a lot of downstream projects running a release quicker than in years previous.

PR #667, opened 2020-06-03
Restore binary compatibility with 1.14.0

smarter:
dbaeb80e5f1e2aad36a647f4fe55fcfde9d1153e broke binary compatibility by
making `BooleanOperators` private, this does not match the plan discussed
in #540:

> The 1.15.x series will preserve binary-compatibility with 1.14.0 but
  is expected to break source-compatibility (e.g. removing the implicit
  keyword from some definitions)

This commit implements that: BooleanOperators is public again but is
also made non-implicit.

### DIFF
--- a/project/MimaSettings.scala
+++ b/project/MimaSettings.scala
@@ -19,7 +19,6 @@ object MimaSettings {
   )
 
   private def removedPrivateMethods = Seq(
-    "org.scalacheck.Prop.BooleanOperators"
   )
 
   private def removedPrivateClasses = Seq(
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -357,7 +357,7 @@ object Prop {
    * values available in the current scope. See [[Prop.ExtendedBoolean]] for
    * documentation on the operators. */
   @deprecated("Please import Prop.propBoolean instead", since="1.14.1")
-  implicit private[this] def BooleanOperators(b: => Boolean): ExtendedBoolean = new ExtendedBoolean(b)
+  def BooleanOperators(b: => Boolean): ExtendedBoolean = new ExtendedBoolean(b)
 
   /** Implicit conversion of Boolean values to Prop values. */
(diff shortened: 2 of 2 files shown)


ashawley:
I wrote that I had made the change in https://github.com/typelevel/scalacheck/pull/498#issuecomment-557582744.  I believe there was a reason for it rather than just removing the implicit.  I'll need to try and remember why I did this.  I admit that my comment was unfortunately very cryptic.

> The `BooleanOperators` is now `private[this]`, for backwards compatibility, in favor of importing the implicit `propBoolean` instead.

non:
@ashawley OK to merge this?

non:
I'm just gonna go ahead and merge this. Thanks again @smarter for catching this!

ashawley:
I remember the motivation now for changing `BooleanOperators` to `private[this]`:  It was for usability around error-messages v. deprecation messages.

ashawley:
Further, changing the implicit to package-private actually is binary compatible.  Mima is flagging it as a binary compatible change, but that is only for Java callers, see https://github.com/lightbend/mima/pull/375.  

It's good that Mima does this for those occasional Scala libraries with downstream Java users.  I'm pretty confident we don't have any Java callers, so this Mima check could be ignored.

smarter:
> Further, changing the implicit to package-private actually is binary compatible. 

Maybe, but before this PR, `private[this]` was used which does get emitted as private and isn't binary compatible.

ashawley:
I believe that would emit the same byte code as the example in https://github.com/lightbend/mima/pull/375, but I could re-check.

smarter:
It doesn't:
```scala
package bla

class A {
  private def foo: Int = 1
  private[bla] def bar: Int = 1
}
```
```java
% scalac A.scala
% javap -p bla/A.class
Compiled from "pp.scala"
public class bla.A {
  private int foo();
  public int bar();
  public bla.A();
}
```

ashawley:
So it does.  I guess I made the incorrect change, then.  I wrongly assumed `private[this]` was like package-private and therefore binary compatible.   The change should have been `private[scalacheck]`.  Thanks.

Issue #664, opened 2020-05-26
Improve choose[BigInt]


ashawley:
`Gen.choose[BigInt]` was added in #636 by @dmurvihill.  [In a comment](https://github.com/typelevel/scalacheck/pull/636#issuecomment-624427814), Erik shared some code that could improve its performance.

Issue #637, opened 2020-03-03
Add Choose[BigDecimal]


dmurvihill:
I was initially going to just do this as part of #636, but now I'm not sure if a `Choose[BigDecimal]` implicit is even a good idea.

Any practical generator of real numbers can never be truly uniform, as `Gen.choose` purports to be. This isn't a problem for `choose[Double]`, `choose[Long]`, or even `choose[BigInt]`, as all three of those values have bounded precision anyway. But when we try to implement `choose[BigDecimal]`, we have to somehow accept a new free variable, precision, that is not available via the `Gen.choose` interface.

One way around this could be: instead of implementing `choose[BigDecimal]`, provide a `Choose[BigDecimal]` factory that accepts a precision and whose result the user could declare as implicit. Seems pretty clunky, but still maybe easier than writing a `BigDecimal` generator directly?

ashawley:
Thanks for creating this issue.  I don't use a lot of arbitrary precision, and you sound like you have a specific use-case in mind, but why would you need to specify precision?  What would be wrong with?

```scala
Gen.choose(BigDecimal("-9.22337203685477580E+214748366"), BigDecimal("9.223372036854775807E-2147483629")) 
```

dmurvihill:
The numbers returned by that generator would be over 500MB in length.

ashawley:
Ok, and that is what a precision parameter you mentioned would help avoid, or is this just an issue with the large numbers I chose as an example?  I assumed `BigDecimal` had an efficient implementation, but this isn't the case?

dmurvihill:
Let's simplify and say we are asking for `Gen.choose(BigDecimal("-1.0"), BigDecimal("+1.0"))`.

Can this generator produce `0.1`? If so, can it produce `0.11`? How about `0.11111111111111111111111111111111111111111111111111111111111111111111`, or even a string of 2^32 1s? Where does it end? That question can only be answered by the user based on their analysis of what is "good enough" for their use case.

ashawley:
Good point. Finding the range between the values would likely determine what the scale should be.  According to the documentation for [java.math.BigDecimal](https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html), subtraction adopts the largest scale of the two operands.  Would that be sufficient?

dmurvihill:
Oh, I really like that. Hang on a minute.

non:
https://github.com/typelevel/scalacheck/pull/670

PR #670, opened 2020-07-01
Add choose methods for both Java and Scala BigDecimal.

non:
In addition to implicit Choose instances for scala.math.BigDecimal and
java.math.BigDecimal we also include explicit constructor methods, since users
may wish to be explicit about the scale they want. We may want to put those
methods directly on Gen, currently the ergonomics of using this are a bit bad:

    Gen.Choose.chooseBigDecimalScale(100).choose(0, 1)

The BigDecimal generation is not yet tested. That will also be added in a
follow up. This also optimizes the BigInt generator a bit and generalizes it to
java.math.BigInteger to support that as well.

Addresses #631, #637, and #664

### DIFF
--- a/src/main/scala/org/scalacheck/Gen.scala
+++ b/src/main/scala/org/scalacheck/Gen.scala
@@ -24,6 +24,7 @@ import scala.concurrent.duration.{Duration, FiniteDuration}
 
 import java.util.{ Calendar, UUID }
 import java.nio.ByteBuffer
+import java.math.{BigInteger, BigDecimal => JavaDecimal}
 
 sealed abstract class Gen[+T] extends Serializable { self =>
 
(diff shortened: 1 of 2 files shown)


src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -449,22 +508,85 @@ object Gen extends GenArities with GenVersionSpecific {
...
+     * See chooseBigDecimalScale for comments.
+     */
+    def chooseJavaBigDecimalScale(minScale: Int): Choose[JavaDecimal] =
+      new Choose[JavaDecimal] {
+        def choose(low: JavaDecimal, high: JavaDecimal): Gen[JavaDecimal] =
+        (low compareTo high) match {
+          case n if n > 0 => throw new IllegalBoundsError(low, high)
+          case 0 => Gen.const(low)
+          case _ => /* n < 0 */
+            val s = (low.scale max high.scale) max minScale
```
ashawley, line 581: It's less explicit, but couldn't the user just specify the minimum desired scale on either of the parameters for `high` and `low` to the `Gen.choose` method?  For example,

```scala
forAll(Gen.choose(BigDecimal(-2).pow(29).setScale(10),
                  BigDecimal( 2).pow(29))) { (d: BigDecimal) =>
  d.scale >= 10
}
```

ashawley: This also came up in the discussion in #637, but I didn't conclude that there's a justification for the `minScale` parameter to the `Choose` factory.

ashawley: You could accomplish this by just changing this line in your pr:

```scala
val s = (low.scale max high.scale) max (low.scale min high.scale)
```

non: I don't understand what that last line accomplishes, the LHS (`low.scale max high.scale`) will always be chosen over the RHS (`low.scale min high.scale`).

My experience here is that many people using `BigDecimal` don't necessarily think about scale. For example, code like 'Gen.choose(BigDecimal(0), BigDecimal(1))` will only produce two possible values, but this issue won't necessarily be detected by authors, leading to more testing.

Using the implicit scale in `BigDecimal` has tripped people up, especially given how tricky it is to reason about. For example, authors might think that `Gen.choose(BigDecimal(0.0000), BigDecimal(1.0000))` will give them a scale of 4, but actually it will be 1 due to a different treatment of `String` and `Double` constructors:

```scala
scala> val x = BigDecimal(0.0000)
val x: scala.math.BigDecimal = 0.0

scala> x.scale
val res0: Int = 1

scala> val y = BigDecimal("0.0000")
val y: scala.math.BigDecimal = 0.0000

scala> y.scale
val res1: Int = 4
```

I'm open to simplifying the API a bit but I strongly feel like we should have a relatively rigorous `minScale`, even if it's hardcoded. It's possible to truncate a generated `BigDecimal` for less precision but the reverse is not possible, so I feel like more precision (to a point) is a better default.

larsrh:
Should we go ahead with this for 1.15.0?

non:
@larsrh I'm happy to including this as-is. I know @ashawley had concerns about it, so I didn't want to merge it without hearing more from him.

larsrh (approved):

ashawley:
I will try to take a look at this again.  I had written some code that experimented with this. Hopefully, it can help me remember where I left off.

non:
@ashawley Ping?

If I made the methods taking `minScale` (i.e. `chooseBigDecimalScale` and `chooseJavaBigDecimalScale`) private for now, would that make you feel better? Then we wouldn't be committed to keeping the other methods around for binary compatibility. Then we could get a release out while giving you more time to consider the issue.

non:
I'm going to go ahead and make the change I suggested -- it seems like a reasonable step to do regardless.

larsrh (approved):

larsrh:
Thanks, Erik!

PR #672, opened 2020-07-17
Make Gen[Tuple] serializable and add regression test

yhylord:
Right now, tuple generators (e.g. `Gen[(String, Int)]`) are not serializable. Generated functions that have tuples in the result (e.g. `List[Int] => List[(String, Int)]`) also don't serialize.
This PR adds serializability to the mentioned values. It does so by removing `Function1` which is not serializable and use `flatMap` and `map` to generate `zip` code in `codegen.scala`. A serialization regression test is also included.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,6 +78,13 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
 
   unmanagedSourceDirectories in Test += (baseDirectory in LocalRootProject).value / "src" / "test" / "scala",
 
+  unmanagedSourceDirectories in Test += {
+    val s = if (scalaMajorVersion.value >= 13 || isDotty.value) "+" else "-"
+    (baseDirectory in LocalRootProject).value / "src" / "test" / s"scala-2.13$s"
+  },
+
+  libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value % "test",
+
   resolvers += "sonatype" at "https://oss.sonatype.org/content/repositories/releases",
 
   // 2.11 - 2.13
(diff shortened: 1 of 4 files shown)


non:
@yhylord It looks like these instances are serializable on 2.13 but are _not_ serializable on 2.11 or 2.12 on the JVM. (We're not running these tests on the JS platform which is why that's OK.)

The change does seem to help 2.13; it's possible for 2.11 and 2.12 that we'd need to create our own serializable `Function1` subclass to get this working.

non:
@yhylord You can choose to build/test Scala 2.12.10 (instead of 2.13) by running `++2.12.10` from SBT before your other commands. For example:

```
$ sbt
... various startup output of SBT ...
sbt:scalacheck> ++2.12.10
[info] Setting Scala version to 2.12.10 on 4 projects.
[info] Excluded 1 projects, run ++ 2.12.10 -v for more details.
[info] Reapplying settings...
[info] set current project to scalacheck (in build file:/Users/erik/w/scalacheck/)
sbt:scalacheck> jvm/compile
[info] Compiling 9 Scala sources to /Users/erik/w/scalacheck/jvm/target/scala-2.12/classes ...
[info] Compiling 8 Scala sources to /Users/erik/w/scalacheck/jvm/target/scala-2.12/classes ...
[success] Total time: 20 s, completed Jul 17, 2020 9:13:27 PM
sbt:scalacheck> jvm/test
... test output follows ...
```

The command is stateful so it will last until you restart SBT. You can use different Scala versions there (e.g. 2.11.12 to test on 2.11 as well).

larsrh:
Rebased & merged in #795.

ashawley:
To be honest, the commit in #795 did credit Haoyang Yu, but it was slightly modified from the original.

ashawley:
The discussion of tuple serialization above last year suggested that this would work on 2.13, but not 2.12.  I couldn't find that to be the case.  I took the change made here and added some test cases to the existing serialization suite and created #795. It worked fine for both Scala versions on the JVM, fwiw.  I'm not sure what changed since last year.  It's not clear what's different about `Function1` from using Scala's function syntax, but it seemed like the right change going forward, regardless.

PR #674, opened 2020-08-03
Fail Seed with (0, 0, 0, 0)

nevillelyh:
Seed(0, 0, 0, 0) would always produce the same next value

### DIFF
--- a/src/main/scala/org/scalacheck/rng/Seed.scala
+++ b/src/main/scala/org/scalacheck/rng/Seed.scala
@@ -99,6 +99,9 @@ sealed abstract class Seed extends Serializable {
 object Seed {
 
   private case class apply(a: Long, b: Long, c: Long, d: Long) extends Seed {
+    if (a == 0 && b == 0 && c == 0 && d == 0) {
+      throw new IllegalArgumentException("illegal Seed(0, 0, 0, 0)")
+    }
     override def toString: String = s"""Seed.fromBase64("$toBase64")"""
   }
 
--- a/src/main/scala/org/scalacheck/rng/Seed.scala
+++ b/src/main/scala/org/scalacheck/rng/Seed.scala
@@ -99,9 +99,6 @@ sealed abstract class Seed extends Serializable {
 object Seed {
 
   private case class apply(a: Long, b: Long, c: Long, d: Long) extends Seed {
-    if (a == 0 && b == 0 && c == 0 && d == 0) {
-      throw new IllegalArgumentException("illegal Seed(0, 0, 0, 0)")
-    }
     override def toString: String = s"""Seed.fromBase64("$toBase64")"""
   }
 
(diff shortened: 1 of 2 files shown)


nevillelyh:
All builds passed except this one due to a time out. Can someone with admin rights re-trigger?
https://travis-ci.org/github/typelevel/scalacheck/jobs/714591093

SethTisue:
green now. (but someone else will have to review the content of the PR)

src/main/scala/org/scalacheck/rng/Seed.scala
```diff
@@ -99,6 +99,9 @@ sealed abstract class Seed extends Serializable {
 object Seed {
 
   private case class apply(a: Long, b: Long, c: Long, d: Long) extends Seed {
+    if (a == 0 && b == 0 && c == 0 && d == 0) {
+      throw new IllegalArgumentException("illegal Seed(0, 0, 0, 0)")
```
dwijnand, line 103: Why not check this in `fromLongs`?

nevillelyh: I actually discovered this via with a [magnolia](https://github.com/propensive/magnolia) derivation of `Arbitrary[T]`. Turns out magnolia can derive the `sealed abstract class Seed` by calling its `private case class apply(...)`.

Derivation code:
https://github.com/spotify/magnolify/blob/master/scalacheck/src/main/scala/magnolify/scalacheck/semiauto/ArbitraryDerivation.scala#L25

This calls `apply()`
```
implicitly[Arbitrary[Seed]].arbitrary.sample
```

Checking in `apply` would prevent these macro libraries from accidentally deriving a bad seed.

dwijnand: Is it calling `private case class apply` or its own subclass of `Seed`?

It shouldn't be able to do either, as `Seed` is `sealed` and `apply` is `private`, so I'd argue that particular behaviour is a problem that needs resolving in Magnolia.  (But I think this PR has merit regardless.)

nevillelyh: Yeah looks like magnolia is able to call the private case class thru macros.

I admit that doing the check here might add a bit of overhead, since `Seed#apply(s: Long)` does this `while (i < 20) { seed = seed.next; i += 1 }` which calls `apply()`.
I can probably override the implicit in my library, but such a check is still nice to have, maybe in `fromLongs` as you suggested?

dbg

non: So, the intention was for this constructor to be private -- the issue of an all-zero seed is known and the code was being defensive to avoid it.

I'm fine adding this check to `fromLongs` but I'd rather not require every (internal) use of `apply` to do the check.

nevillelyh: Cool, my thought as well. Will move it to `fromLongs` and fix it on my end to avoid `apply(0, 0, 0, 0)`.

nevillelyh: @non moved to `fromLongs` PTAL thanks :)

non:
LGTM 👍

PR #677, opened 2020-08-21
Prevent nesting properties

NthPortal:
Prevent nesting properties or otherwise adding properties
during execution.

### DIFF
--- a/src/main/scala/org/scalacheck/Properties.scala
+++ b/src/main/scala/org/scalacheck/Properties.scala
@@ -29,6 +29,7 @@ import util.ConsoleReporter
 class Properties(val name: String) {
 
   private val props = new scala.collection.mutable.ListBuffer[(String,Prop)]
+  private var frozen = false
 
   /**
    * Customize the parameters specific to this class.
@@ -46,7 +47,10 @@ class Properties(val name: String) {
 
   /** Returns all properties of this collection in a list of name/property
    *  pairs.  */
-  def properties: collection.Seq[(String,Prop)] = props
+  def properties: collection.Seq[(String,Prop)] = {
+    frozen = true // once the properties have been exposed, they must be frozen
+    props
+  }
 
   /** Convenience method that checks the properties with the given parameters
    *  (or default parameters, if not specified)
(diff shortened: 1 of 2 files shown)


src/main/scala/org/scalacheck/Properties.scala
```diff
@@ -46,7 +47,10 @@ class Properties(val name: String) {
 
   /** Returns all properties of this collection in a list of name/property
    *  pairs.  */
-  def properties: collection.Seq[(String,Prop)] = props
+  def properties: collection.Seq[(String,Prop)] = {
+    frozen = true // once the properties have been exposed, they must be frozen
```
NthPortal, line 51: an alternative would be to add a `private[scalacheck] def freeze(): Unit` and call it from the task when it gets the `properties` or right before execution

SethTisue:
does anyone object to this? if not, let's merge it (in a week or so, say)

NthPortal:
I don't see any objections

SethTisue:
okay, if you squash it, I'll merge it

NthPortal:
squashed

ashawley:
What was the motivation for this change?  I recall discussing in Gitter that there was an effort to optimize the Scala collections library that revealed a defect in this code in ScalaCheck.  Is there an upstream issue to read more about this?

NthPortal:
There were two motivations for this change:

- nesting properties was never supported, and it is better to explicitly disallow things you don't support rather than have them partially or mostly work and then have behaviour accidentally change underneath people if you end up changing the internal implementation
- what happens if you nest properties is that it creates a new property while executing the rest of the properties, and adds it to the `ListBuffer` that's being iterated over in order to execute all of the properties. as of 2.13.4, it is very likely that `ListBuffer` iterators (and soon other types as well) will throw an exception if you modify the underlying collection while using them

ashawley:
Thanks, for the explanation.  Had you run across a test suite that had nested properties and an exception was thrown with the `ListBuffer` change?

NthPortal:
yes. scala/scala's 😬

ashawley:
Ok, I wonder if I was wrong about nested properties not being supported.  I know nesting properties isn't documented, but I suppose it may be possible for people to use it.  I recall that cats and discipline projects could be doing something like this with law checking.

SethTisue:
@ashawley would you like me to run this through the Scala community build?

ashawley:
Yes, that would be useful.  I was planning to ask. I'll open an issue there.

SethTisue:
We're good over at https://github.com/scala/community-build/issues/1226, so it looks like this change will stand.

ashawley:
Yes, thanks for doing that for us, Seth!

PR #683, opened 2020-10-08
annotate Prop and Pretty as `@FunctionalInterface`

SethTisue:
this makes them more convenient to use from Dotty, since Dotty warns on SAM types unless they have this annotation

### DIFF
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -22,6 +22,7 @@ sealed class PropFromFun(f: Gen.Parameters => Prop.Result) extends Prop {
 }
 
 @Platform.EnableReflectiveInstantiation
+@FunctionalInterface
 sealed abstract class Prop extends Serializable { self =>
 
   import Prop.{Result, True, False, Undecided, provedToTrue, mergeRes}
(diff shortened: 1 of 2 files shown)


ashawley (approved):

dwijnand:
What the code look like that Dotty warns on?

SethTisue:
> What the code look like that Dotty warns on?

Gah, I'm glad you were a little skeptical... the code in question turned out to be erroneous, and at the time I opened this PR, I hadn't yet seen that it had to do with Dotty handling an opening-brace-at-beginning-of-line differently than Scala 2 does.

So, although this PR might not be outright wrong, it might not be helpful either, and I don't know whether, on the balance, it's likelier to make error message on bad code more helpful or less helpful.

So I'm going to close this unless someone who knows the ScalaCheck API better than I does steps up and says that they think it actually should be merged

dwijnand:
My guess was this was around eta-expanding Prop's/Pretty's apply method, when passed as values, which I (1) didn't know which ScalaCheck API would require that and (2) would be a bad/unfortunate breakage if that were coming from Dotty.

So it sounds like it's a false positive error message by Dotty when significant whitespace is off (which sounds even harder to fix...)

PR #695, opened 2020-11-02
Fix Gen.pick when used with Gen[_] arguments.

non:
Previously, when called with generators Gen.pick could return collections of
any length, when the API promises that they are length `n`.

    // should only generate Seq(0,1), Seq(0,2), or Seq(1,2).
    val g = Gen.pick(2, Gen.const(0), Gen.const(1), Gen.const(2))

We didn't have a test for this variant of Gen, which explains why we didn't
detect this breakage.

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -289,6 +289,19 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
     }
   }
 
+  property("pick with gen") = forAll { (x: Int, y: Int, rest: List[Int]) =>
+    val lst = x :: y :: rest
+    forAll(choose(-1, 2 * lst.length)) { n =>
+      val gs = rest.map(Gen.const)
+      Try(pick(n, Gen.const(x), Gen.const(y), gs: _*)) match {
+        case Success(g) =>
+          forAll(g) { m => m.length == n && m.forall(lst.contains) }
+        case Failure(_) =>
+          Prop(n < 0 || n > lst.length)
+      }
+    }
+  }
+
   /**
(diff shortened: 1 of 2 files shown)


non:
(I'm not sure how widely this variant of `pick` is used but I think it's worth considering releasing 1.15.1 fairly quickly.)

jvm/src/test/scala/org/scalacheck/GenSpecification.scala
```diff
@@ -289,6 +289,19 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
     }
   }
 
+  property("pick with gen") = forAll { (x: Int, y: Int, rest: List[Int]) =>
+    val lst = x :: y :: rest
+    forAll(choose(-1, 2 * lst.length)) { n =>
+      val gs = rest.map(Gen.const)
+      Try(pick(n, Gen.const(x), Gen.const(y), gs: _*)) match {
+        case Success(g) =>
+          forAll(g) { m => m.length == n && m.forall(lst.contains) }
```
ashawley, line 298: Does this assertion need to be `m.forall(x :: y :: lst.contains)`?

ashawley: That has a syntax error, but hopefully you catch my drift.

non: No because `val lst = x :: y :: rest`.

larsrh (approved):

jvm/src/test/scala/org/scalacheck/GenSpecification.scala
ashawley: So it is.  I read `lst` as `rest`.

Issue #704, opened 2020-11-03
Migrate off of Travis CI


non:
It seems like Travis CI has majorly reduced their capacity for building open source projects. We should migrate somewhere else.

ashawley:
I've added an initial shell of a build in GitHub actions in 2163aa4, just so we have a working CI.  

I need to audit it further, but it looks like it builds the primary stuff.  Still need to see how to do these:

- Scala native builds
- Build example projects
- Running benchmark smoke test

ashawley:
For now, maybe we should shut off pull request builds on Travis to avoid queuing builds?  It's good to have when merges to master happen since it's the full build.  I'd do it myself, but I don't have the privs to do it for some reason. Maybe someone else does?  https://travis-ci.org/github/typelevel/scalacheck/settings

![Screen Shot 2020-11-03 at 10 40 19 PM](https://user-images.githubusercontent.com/358615/98066649-9f4a4500-1e25-11eb-986e-d600c31820f7.png)

SethTisue:
fwiw, the repos I moved from travis-ci.org to travis-ci.com recently have all been doing just fine the last few days (jobs run promptly), so at least as a short term measure, someone with admin access in the typelevel org (it's apparently not enough to have admin access just in this repo) could flip the switch to move to travis-ci.com

ashawley:
Ok, good to know.  That would be good, because we don't have concurrency either, so it's a slow parade.

ashawley:
Oh, geez.  Daniel has made a bunch of headway with GitHub Actions in #705.

larsrh:
I just disabled the Travis CI build, so we can technically close this. FWIW, I prefer Daniel's approach with managed workflow configuration.

ashawley:
Ok.  Hopefully, we can better coordinate work here in the future so we avoid duplication.

ashawley:
I was interested in having both CI still working to some degree to sanity check the new setup, but if we're confident that it's fine, then we can shut off Travis.

I guess the only other concern is getting 1.15.1 out, since there are some significant defects that need to go out pretty soon.

PR #705, opened 2020-11-04
Attempting to move the build to github actions

djspiewak:


djspiewak:
BTW I *think* this is ready if anyone wants to make a decision on it.

larsrh:
Wait a second, why is there a conflicting GitHub action in master, currently? @ashawley?

larsrh (approved):
Needs a rebase to account for the workflow scripts already in master, but other than that, I think this is the best way forward.

ashawley:
Yes, since Travis is borked for us, I started working on GitHub Actions going at the same time as Daniel.  See discussion in #704.  I think the build matrix could get paired down here, but otherwise this seems like a good approach.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,254 @@
...
+        run: |
+          cd examples
+          for d in */ ; do cd "$d" && sbt test:compile && cd ../ ; done
+
+  bench:
+    name: Bench
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        scala: [0.27.0-RC1, 3.0.0-M1, 2.11.12, 2.12.10, 2.13.3]
```
middilman, line 200: Probably fine to just smoke test the benchmark on a single Scala version like the examples do.

ashawley, line 200: Probably fine to just smoke test the benchmark on a single Scala version like the examples do.

ashawley (approved):

ashawley:
Yeah, feel free to just kill the files I had added to master directly when rebasing.

project/plugin.sbt
```diff
@@ -15,19 +15,21 @@ def printAndDie(msg: String): Nothing = {
 // Update SCALAJS_VERSION in release.sh, as well
 val scalaJSVersion = env("SCALAJS_VERSION") match {
   case Some("0.6.33") => "0.6.33"
-  case Some("1.3.0") | None => "1.3.0"
+  case Some("1.3.0") | Some("") | None => "1.3.0"
```
ashawley, line 18: Gosh, Erik wanted to have the build script be fatal if the wrong version was supplied, but I suppose that never worked.

Maybe we should delete the `printAndDie`, below, since it doesn't do anything.

djspiewak: Well, if the wrong version is specified, it *is* fatal. It's just that some platforms don't distinguish between an empty envar and an unspecified one.

.github/workflows/ci.yml
djspiewak: Alrighty

project/plugin.sbt
ashawley: Yes, I misread it.  You're right, it's fine.

non:
Is this ready to merge? If we only need minor tweaks I'd prefer to merge and handle those in a follow-up.

larsrh:
I ran out of time today, but it's on my list for tomorrow.

djspiewak:
It's ready to merge IMO. The only tweaks it needs are to poke some holes in the matrices for examples and benchmarks.

Issue #710, opened 2020-11-06
Fragile test of SeedSpecification


ashawley:
I saw a curious failure on GitHub Actions that I hadn't seen before:
```
[info] failing seed for Seed.longs are evenly-distributed is -aQhCwV8oBUYlWwee46hAk7cDCysDENMKQKEw-FnwEC=
[info] ! Seed.longs are evenly-distributed: Falsified after 368 passed tests.
[info] > ARG_0: Seed.fromBase64("mngJYod-N9mPAhd861EV4YJQcckrt-vaCc8Eabb013H=")
[info] > ARG_1: Base(53)
[error] Failed: Total 8, Failed 1, Errors 0, Passed 7
[error] Failed tests:
[error]  org.scalacheck.rng.SeedSpecification
```

ashawley:
Draft PR #711 shows that the seed fails.

I believe this is a probabilistic property test, so it's probably acceptable to fail rarely.

non:
I re-read the test. I think the predicate has a random 1-in-1.7M chance of failing each time we run it, but since we evaluate the `forAll` 100 times, we really have a 1-in-1700 chance of failing each time we run `SeedSpecification`. Since we tend to run CI over a lot of different targets this means we have more like a 1-in-170 (or so) chance of failing.

We can reduce the chances of false positives at the expense of test accuracy: right now we expect the random value we produce to be within 5 standard deviations of the expected result (https://en.wikipedia.org/wiki/Standard_deviation#Rules_for_normally_distributed_data).

If we used 6 standard deviations instead of 5, we'd expect the test to randomly fail only 1-in-506.8M (about 300x less often).

Do you think we should change it? I'm fine closing your PR and this issue, or making that change (which could obscure real bugs but will also cut down on noise).

ashawley:
I was just wondering if your probability test had a flaw or if the the issue was with the random number generator.  A failure did seem to be super rare over the last 4 years as you described.  I'm fine with closing this.

PR #716, opened 2020-11-13
Test the distribution of full long in seed

ashawley:
While looking at the test failure of #710, I came across this line in the test.  Is there a reason not to test the full entropy of the `Seed.long`?  Maybe constraining only the bottom 31 bits is intended?

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/rng/SeedSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/rng/SeedSpecification.scala
@@ -44,7 +44,7 @@ object SeedSpecification extends Properties("Seed") {
       def countZeros(s0: Seed, i: Int, seen0: Int): Int =
         if (i <= 0) seen0 else {
           val (x, s1) = s0.long
-          val n = (x & 0x7fffffff).toInt % base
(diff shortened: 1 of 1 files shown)


SethTisue:
is there any indication in the Git history?

ashawley:
No, not from what I could tell.  Our entropy expert, Erik, will need to weigh in.

non (approved):
This is a good change. 👍 

To try to explain what was going on (at least as far as I can remember):

The `%` operator isn't guaranteed to return positive values, so if you want to be sure to limit the results of `x % base` to the interval `[0, base)` you have to ensure that `x` is non-negative to start with. If the `x` is negative the results are instead in the interval `(-base, 0]`.

For `Int` values you can do this with `x & 0x7fffffff` (aka `x & Int.MaxValue`). That's what I did in this case. For long values you instead need to say `x & 0x7fffffffffffffffL` (aka `x & Long.MaxValue`), which is what I probably should have said based on my thinking.

However, since we're only comparing the results against zero we don't care of the output is positive or negative, meaning the whole exercise was unnecessary. That's why this is a good PR.

PR #717, opened 2020-11-13
Integer.max not needed in Gen.buildableOf

ashawley:
I believe `Gen.sized` is always non-negative, so this check isn't necessary.


rossabaker (approved):
I think that's right. It rejects test parameters with a minSize or maxSize < 0.

ashawley:
I thought it was, too.  However, there's a way to use `Gen.resize` with a negative value.  I've been meaning to circle back on this and add some property tests so we know that users can do that.  Whether that's intentional or not is a different question.

PR #712, opened 2020-11-06
Support for DottyJS

djspiewak:
I've also released 1.15.1 against SJS1 with these changes.

PR #724, opened 2020-11-25
In with 3.0.0-M2, out with 0.27.0-RC1

rossabaker:


rossabaker:
I don't know anything about scala.js.  This built for me locally without propagating this to 3.0.0-M2.  Can we do without it?

```scala
    scalacOptions ++= {
      if (scalaVersion.value == "3.0.0-M1") Seq("-Yskip:explicitJSClasses") else Nil
    },
```

SethTisue:
@djspiewak you added that thing in #712, do you know the answer?

djspiewak:
That was a workaround for a bug in M1. I think it was fixed.

Issue #351, opened 2017-08-02
add more maintainers


xuwei-k:

rickynils:
I'm definitely interested in this.

These are some ideas I currently have, that are more or less related to maintenance:

* Create a `scalacheck` GitHub organization, to emphasize shared maintenance
* Keep the current ScalaCheck lib slim (or make it even slimmer) and dependency-less. Encourage utility libraries for specialized functionality/combinators/generators etc. This does not imply that the ScalaCheck core should stop evolve/improve.
* Improve the release process and versioning scheme. I have some ideas for making it more declarative, which should help maintainers releasing new versions.

I'm open for any kinds of comments and suggestions regarding project maintenance and maintainers. I don't have a lot of time for ScalaCheck work for the moment, but I'm willing to put in some time in order to achieve a better situation that will save both me and the general Scala community time in the future, and that will allow ScalaCheck to evolve further.

cc @non

nrinaudo:
Not sure this is the right place to suggest this, but could one of these utility libraries you speak of be laws for `Cogen`, etc... to allow downstream libraries to more easily validate their implementstions?

An example is the enumeratum-scalacheck module that was published recently that basically copy / pasted tests defined here.

BennyHill:
@rickynils  I hope this suggestion does not pre-empt the way forward, but do you think it might be worth setting up the aforementioned GH organization, lest it is misappropriated?

SethTisue:
this has come up again at #374. @etorreborre asked me for my take there, I guess as a representative of the Scala team at Lightbend?

for our own needs in the scala/scala repo, we can re-in-source, or fork, or whatever.

but for the community and ecosystem, there is also some renewed urgency now that Scala 2.13.0-M4, -M5, and -RC1 are all looming in the coming weeks and months.

making a scalacheck organization sounds right to me, and my guess is that if Rickard says he will authorize volunteer maintainers (and, more to the point for the 2.13 need, releasers) and asks them to step forward, they will.

rickynils:
As you've all noticed, I have more or less gone offline for the last year. This is a temporary situation, and I very much like to keep ScalaCheck evolving. I do, however, wish for ScalaCheck to be less attached to me and more able to stand on its own (which is what this issue is about).

My immediate plan is to release the current master branch as 1.14.0 (#374), and to release any Scala milestone builds that is needed, but hopefully get a team of maintainers up and running (and able to make releases) for the releases after that.

[...]

PR #734, opened 2020-12-18
Add java.time._ Instances For The JVM

isomarcte:
This commit adds Scalacheck instances for almost all of the types in the `java.time` package and sub-packages, for which meaningful definitions seem to exist.

Almost all implemented types have `Gen.Choose` instances. `Shrink` instances were defined for the types that seemed to have a reasonably straight forward definition. And all implemented types have an `Arbitrary` instance.

The instances were built against JRE 8, though I'm not aware of any changes between the JRE definitions and the current (JRE 15) definitions.

The types are scoped to `org.scalacheck.time`, with the intent that `import org.scalacheck.time._` would bring them all into the implicit scope.

It is likely worth discussing why these types are being added. Scalacheck instances for `java.time` types are hardly new. There are several libraries which provide some instances for some `java.time` types. There is however, no library of which I am aware, which provides a relatively complete implementation of these types, e.g. instances for most types, with full range support, and `Choose` instances. If such a library already exists, then I am more than happy to withdraw this PR.

Further, `java.time` types have a relatively large user base in the Scala community. By hosting instances in Scalacheck proper, we increase exposure to these instances and lessen the likelihood that developers will attempt to re-invent the wheel, or just use stub instances, e.g. `Gen.const(LocalDate.now)`.

While these types are currently scoped to the JVM only, it is arguable that they could be provided for ScalaJs and ScalaNative too. Both ScalaJs and ScalaNative have, or are working on, replacement API compatible implementations. However adding support in Scalacheck proper for this would require adding dependencies and is thus punted to a later discussion for now.

### DIFF
--- a/jvm/src/main/scala/org/scalacheck/time/JavaTimeInstances.scala
+++ b/jvm/src/main/scala/org/scalacheck/time/JavaTimeInstances.scala
@@ -1,0 +1,556 @@
+package org.scalacheck.time
+
+import org.scalacheck._
+import java.time._
+import java.time.temporal._
+import org.scalacheck.Gen.Choose
+
+/** Instances for `java.time` types. */
+private[time] trait JavaTimeInstances {
+
+  // temporal.ChronoUnit
+  //
+  // Arbitrary and Choose instances are already provided by Java enum support.
+
+  implicit final lazy val cogenChronoUnit: Cogen[ChronoUnit] =
+    Cogen[Int].contramap(_.ordinal)
+
+  // Duration
+
+  // Java duration values are conceptually infinite, thus they do not expose
+  // Duration.MAX/Duration.MIN values, but in practice they are finite,
+  // restricted by their underlying representation a long and an int.
+
+  private final lazy val minDuration: Duration =
+    Duration.ofSeconds(Long.MinValue)
+
+  private final lazy val maxDuration: Duration =
+    Duration.ofSeconds(Long.MaxValue, 999999999L)
+
+  implicit final lazy val chooseDuration: Choose[Duration] =
+    new Choose[Duration] {
+      override def choose(min: Duration, max: Duration): Gen[Duration] = {
+        min.compareTo(max) match {
+          case 0 => Gen.const(min)
+          case result if result > 0 => Gen.fail
+          case _ =>
+            val minSeconds: Long = min.getSeconds
+            val maxSeconds: Long = max.getSeconds
+            Gen.choose(minSeconds, maxSeconds).flatMap{seconds =>
+              val minNanos: Int =
+                if(seconds == minSeconds) {
+                  min.getNano
+                } else {
+                  1
+                }
+              val maxNanos: Int =
+                if (seconds == maxSeconds) {
+                  max.getNano
+                } else {
+                  999999999
+                }
+              Gen.choose(minNanos, maxNanos).map(nanos =>
+                Duration.ofSeconds(seconds, nanos.toLong)
+              )
+            }
+        }
+      }
+    }
+
+  implicit final lazy val arbDuration: Arbitrary[Duration] =
+    Arbitrary(Gen.choose(minDuration, maxDuration))
+
+  implicit final lazy val cogenDuration: Cogen[Duration] =
+    Cogen[(Long, Int)].contramap(value => (value.getSeconds, value.getNano))
+
+  implicit final lazy val shrinkDuration: Shrink[Duration] =
+    Shrink[Duration]{value =>
+      val q: Duration = value.dividedBy(2)
+      if (q == Duration.ZERO) {
+        Stream(Duration.ZERO)
+      } else {
+        q #:: q.negated #:: shrinkDuration.shrink(q)
+      }
+    }
+
+  // Instant
+
+  implicit final lazy val chooseInstant: Choose[Instant] =
+    new Choose[Instant] {
+      override def choose(min: Instant, max: Instant): Gen[Instant] =
+        min.compareTo(max) match {
+          case 0 => Gen.const(min)
+          case result if result > 0 => Gen.fail
+          case _ =>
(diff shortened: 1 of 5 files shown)


isomarcte:
I've encoded trait which brings the implicits into scope as such,

```scala
package object time extends time.JavaTimeInstances
```

Though I acknowledge that several other popular encodings exist. For example,

```scala
package object time {
    object implicits extends time.JavaTimeInstances
}
```

or

```scala
package object time {
    object syntax {
      object all extends time.JavaTimeInstances
    }
}
```

If the maintainers have any preference's here, I'm more than happy to make changes to that structure.

isomarcte:
Also totally happy to move these if they would be better in another module or to close the PR if they aren't wanted. :)

isomarcte:
Looks like Scala is missing an `Ordering` for some types on 2.11...I'll look into that.

isomarcte:
Fixed the compilation issues on Scala < 2.13. It was having issues resolving `Ordering` instances for three of the `java.time` types.

ashawley:
This is a nice contribution.  Would a better home for this be https://github.com/47degrees/scalacheck-toolbox though?

isomarcte:
@ashawley So that's a bit of a tricky question. I considered proposing this PR there, but I decided not to do so for these reasons,

* https://github.com/typelevel/scalacheck already implements instances for other core JVM types. For the purposes of exposure to the user base, it seemed to me that it would make most sense in this repo. Especially since it doesn't require any new dependencies.
* https://github.com/47degrees/scalacheck-toolbox already contains a subset of instances for the types in this PR, but they are implemented in a radically different way. In order to add these to that repo, it would effectively replace the entirety of the implementation there.
* That codebase also deals with JodaTime. Any contributions like this to that code base might reasonably expect to have similar Joda implementations.
* Adds new typeclasses on top of their implementation, which are required in the interface, e.g. `Granularity`. There's nothing objectively wrong with that, but it is nice to just have a core implementation in terms of the JRE APIs and Scalacheck. Other items could be built on top of that in other libraries.

isomarcte:
Anything I can do to move this along? Or should I close it? Or do people just need more time to think it over?

larsrh:
I'm leaning towards inclusion. @ashawley, do you veto it?

SethTisue:
is there somebody at 47 Degrees we should solicit an opinion from? (@alejandrohdezma, perhaps?)

ashawley:
FWIW, there's also this library: https://github.com/tmccarthy/intime

ashawley:
@larsrh No, I wasn't trying to veto.  I was attempting, but wasn't convincing enough, to facilitate getting this contributed to a library elsewhere in the ScalaCheck ecosystem.  In the spirit of what Richard wrote at https://github.com/typelevel/scalacheck/issues/351#issuecomment-322763942 about keeping the library minimal and because we don't have a lot of capacity for maintaining the project, we should probably lean towards leaning the capabilities of the library to some definition, like Scala its standard library and some Java primitives.

SethTisue:
cc @tmccarthy

tmccarthy:
Just providing my humble opinion since my library was tagged above.

This seems like a nice addition that provides basically the same functionality as the [`intime-scalacheck`](https://github.com/tmccarthy/intime#intime-scalacheck) library I wrote. Obviously I'm not sufficiently involved in this project to weigh in on the balance between maintaining additional instances vs contributing this to another library. But I'd note that scalacheck already provides [instances for the `java.util.Calendar` class](https://github.com/typelevel/scalacheck/blob/2224f85/src/main/scala/org/scalacheck/Gen.scala#L1332), which is _de facto_ deprecated. Given that the `java.time` classes are today the standard representation of datetime concepts, it seems appropriate for instances to be provided as part of the main scalacheck library.

Probably the only value `intime-scalacheck` would provide over the instances in this PR is what I've termed ["sensible generators"](https://github.com/tmccarthy/intime/blob/619a4e0/scalacheck/src/main/scala/au/id/tmm/intime/scalacheck/SensibleJavaTimeGenerators.scala). I've found these more useful than using `Arbitrary` instances in many cases. But I think it would be reasonable for something like that to be provided by an external library.

larsrh (approved):
Thanks @tmccarthy for chiming in and @isomarcte for the contribution. I'm a friend of optimistic merging, so I'm suggesting to do that here. If noone objects, I'll push the merge button in a few days.

ashawley:
If contributing this to other ScalaCheck libraries isn't ideal, then adding to ScalaCheck core seems fine.  Time values aren't Java primitives, but they are used enough that it seems like a valid exception.

ashawley:
> The types are scoped to `org.scalacheck.time`, with the intent that `import org.scalacheck.time._` would bring them all into the implicit scope.

This is very smart, but I would rather they just be added to the library the same way as other types.  I don't think we need to add the namespace.  It might be something to consider in the future.

isomarcte:
@ashawley just to clarify, are you saying you'd like the instances to be mixed into the companion objects for the various typeclasses, e.g. `object Arbitrary extends ArbitraryLowPriority with ArbitraryArities with JavaTimeArbitraries` or similar?

If so, do you want me to remove the `org.scalacheck.time` package entirely and just scope everything to `org.scalacheck` or should I keep the actual `private[scalacheck] trait JavaTimeArbitraries` scoped to `org.scalacheck.time`?

ashawley:
Since it's private and if it doesn't get in the way, then I don't see any reason not to use the `org.scalacheck.time` namespace if you wish to.

isomarcte:
@ashawley I'll update it to reflect those changes.

isomarcte:
@ashawley @larsrh the latest commit splits up the `JavaTimeInstances` trait into 4 traits and has the companion object for each typeclass extends the corresponding trait. Let me know if you'd like me to make any further changes.

larsrh:
Thanks a ton @isomarcte!

PR #740, opened 2020-12-22
Enable customization of `Commands.Command` shrinking

jonaskoelker:
The `Commands` user can either specify a `Shrink[Command]`, in which
case the default `List` shrinker will be used to shrink sequential and
parallel commands, or the user can specify those shrinks directly, or
a mix (e.g. `Shrink[Command]` and `Shrink[List[List[Command]]]`).

### DIFF
--- a/src/main/scala/org/scalacheck/commands/Commands.scala
+++ b/src/main/scala/org/scalacheck/commands/Commands.scala
@@ -262,6 +262,24 @@ trait Commands {
     * [[State]].  By default no shrinking is done for [[State]]. */
   def shrinkState: Shrink[State] = implicitly
 
+  /** Override this to provide a custom Shrinker for your internal
+    * [[Command]].  By default no shrinking is done for [[Command]]. */
+  def shrinkCommand: Shrink[Command] = implicitly
+
+  /** Override this to provide a custom Shrinker of sequential [[Command]]s.
+    * By default, the implicit List shrinker is used with [[shrinkCommand]]. */
+  def shrinkSequentialCommands: Shrink[List[Command]] = {
+    implicit val commandShrinker = shrinkCommand
+    implicitly
+  }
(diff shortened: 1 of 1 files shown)


jonaskoelker:
This is one part of https://github.com/typelevel/scalacheck/pull/632, split out into its own PR. See also https://github.com/typelevel/scalacheck/pull/739.

jonaskoelker:
I'd be happy to add documentation and tests; however, I'm not sure exactly what your documentation requirements are, and I think the tests belong in the same file as the one created by #739, and that seems like the least fun kind of merge commit :-) so let's land #739 first?

ashawley:
Can you define an `implicit val shrinkActions` in the`Properties` test class to replace the one defined in `Commands` and have complete control of the shrinking of actions and thus commands?

jonaskoelker:
> Can you define an `implicit val shrinkActions` in the`Properties` test class to replace the one defined in `Commands` and have complete control of the shrinking of actions and thus commands?

The main obstacle is that the `case class Actions` is private. That can be changed. So is `shrinkActions` and that can be changed too. However, I think one of the value propositions of the `Commands` module is that it does the command sequence shrinking for me—I as a user don't have to implement shrinking of sequences of commands myself, only of individual commands (and states). If I override `shrinkActions` I'll have to reimplement the shrinking of command sequences.

Maybe what you had in mind is making `shrinkActions` customizable, but making the default implementation available, maybe with state and command shrinking customizable? But isn't that exactly what this PR does, except it doesn't make `shrinkActions` overridable? If what you want is this PR plus a customizable `shrinkActions`, I think the best path is via two PRs: this one, plus one that makes `shrinkActions` customizable.

If you insist I'll make a PR which just overrides `shrinkActions`, but please consider whether that's really what you want.

PR #744, opened 2020-12-30
Shrink all tuple sizes

jonaskoelker:
This shrinks tuples of all sizes between 2 and 22, both inclusive.

Today I learned you do automatic code generation for various bits that are symmetric across many arities. That seems like a natural and good thing to use for tuple shrinking.

I added a few test cases. With the current tuple shrinking `eqvTupleShrinks(xs, ys)` could just return `xs == ys`. However, I think tuple shrinking might be better if we interleave rather than append the various single-component streams, and I'm pretty sure the `.sorted` equivalence holds if we make that change—that is, I'm deliberately testing something weaker than what I know to be true, in order to future-proof.

Would you like me to make the interleaving change? The advantage of doing e.g. a snake draft across all non-empty single-component shrink streams is that when you shrink component k of an n-tuple, in a situation where no shrunk values of the first k-1 components can provoke a test failure, the number of elements you need to look at before trying to shrink component k is k-1 rather than $sum_{i to k-1} number-of-shrinks-of-component-i$. Stated another way: relatively speaking it prioritizes earlier shrinks of later components rather than later shrinks of earlier components.

If you think there are advantages to prioritizing focusing on the first component, I'll be happy to hear about them. (One could also do a hybrid where you foldLeft a pairwise interleaving, such that the complete interleaving gives 2^-k of the attention to the k'th component.)

### DIFF
--- a/project/codegen.scala
+++ b/project/codegen.scala
@@ -50,6 +50,27 @@ object codegen {
   def fntype(i: Int) =
     s"(${types(i)}) => Z"
   
+  def shrinkTuple(i: Int): String =
+    s"""|  /** Shrink instance of ${i}-tuple */
+        |  implicit def shrinkTuple${i}[
+        |    ${shrinkTypes(i)}
+        |  ]: Shrink[(${types(i)})] =
+        |    Shrink { case (${vals(i)}) =>
+        |      ${shrinkEachComponentOneByOne(i)}
+        |    }""".stripMargin
+
+  def shrinkTypes(n: Int): String =
+    (1 to n).map(i => s"T${i}:Shrink").mkString(", ")
+
+  def shrinkEachComponentOneByOne(n: Int): String =
+    (1 to n).map(shrinkSingleComponent(n, _)).mkString(" append\n      ")
+
+  def shrinkSingleComponent(n: Int, i: Int): String =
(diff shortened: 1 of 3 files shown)


satorg:
@jonaskoelker , sorry for the long response. This PR slipped unnoticed somehow.
I feel however, it makes sense to have the proposed functionality included into the library.

Would you be open to actualize your PR against the current `main` please (i.e. resolve the conflicts and so on)?
Thank you in any case!

project/MimaSettings.scala
```diff
@@ -25,6 +25,14 @@ object MimaSettings {
...
   private def otherProblems = Seq(
-  )
+    "org.scalacheck.Shrink.shrinkTuple2",
+    "org.scalacheck.Shrink.shrinkTuple3",
+    "org.scalacheck.Shrink.shrinkTuple4",
+    "org.scalacheck.Shrink.shrinkTuple5",
+    "org.scalacheck.Shrink.shrinkTuple6",
+    "org.scalacheck.Shrink.shrinkTuple7",
+    "org.scalacheck.Shrink.shrinkTuple8",
+    "org.scalacheck.Shrink.shrinkTuple9"
```
satorg, line 35: What makes the new generated methods incompatible with the current ones, could you clarify please? At first glance, the generated method should be very similar to those that they are supposed to replace.

PR #752, opened 2021-01-22
Scala Native 0.4.0

gabro:
Patch to #751 to make it compile, it still crashes due to reflective instantiation of the test classes.

<details>
<summary>Stack trace</summary>

[error] scala.scalanative.testinterface.common.RPCCore$RPCException: java.lang.Exception: loadable module not found: org.scalacheck.NoPropertyNestingSpecification
[error]         at scala.scalanative.testinterface.common.RPCCore.$anonfun$handleMessage$4(RPCCore.scala:66)
[error]         at scala.Option.foreach(Option.scala:407)
[error]         at scala.scalanative.testinterface.common.RPCCore.$anonfun$handleMessage$1(RPCCore.scala:64)
[error]         at scala.scalanative.testinterface.common.RPCCore.$anonfun$handleMessage$1$adapted(RPCCore.scala:43)
[error]         at scala.scalanative.testinterface.common.Serializer$.withInputStream(Serializer.scala:51)
[error]         at scala.scalanative.testinterface.common.RPCCore.handleMessage(RPCCore.scala:43)
[error]         at scala.scalanative.testinterface.NativeRunnerRPC.$anonfun$runner$1(NativeRunnerRPC.scala:29)
[error]         at scala.scalanative.testinterface.NativeRunnerRPC.$anonfun$runner$1$adapted(NativeRunnerRPC.scala:29)
[error]         at scala.scalanative.testinterface.ComRunner$$anon$1.run(ComRunner.scala:60)
[error] Caused by: java.lang.Exception: loadable module not found: org.scalacheck.NoPropertyNestingSpecification
[error]         at java.lang.Throwable.fillInStackTrace(Unknown Source)
[error]         at org.scalacheck.Platform$.$anonfun$loadModule$1(Unknown Source)
[error]         at org.scalacheck.Platform$$$Lambda$1.apply(Unknown Source)
[error]         at scala.Option.getOrElse(Unknown Source)
[error]         at org.scalacheck.Platform$.loadModule(Unknown Source)
[error]         at org.scalacheck.ScalaCheckRunner.rootTask(Unknown Source)

</details>

I'm a bit puzzled, since it looks like the test classes seem be erased at runtime, despite their superclass `Properties` being annotated with `@EnableReflectiveInstantiation`

### DIFF
--- a/native/src/main/scala/org/scalacheck/Platform.scala
+++ b/native/src/main/scala/org/scalacheck/Platform.scala
@@ -11,7 +11,7 @@ package org.scalacheck
 
 import Test._
 
-import scala.scalanative.testinterface.PreloadedClassLoader
+import scala.scalanative.reflect.Reflect
 
 private[scalacheck] object Platform {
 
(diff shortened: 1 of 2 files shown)


gabro:
Asked about it here https://gitter.im/scala-native/scala-native?at=600a8bf1753011449b0bfbf8

larsrh:
Wild speculation: maybe because most ScalaCheck tests are actually `object`s and not `class`es? Does the annotation also get applied to `object`s inheriting from `Properties`?

gabro:
@larsrh It should. The `loadModule` API is specifically for loading module classes (aka objects).

FWIW it also fails with classes.

Notice that manually annotating all test classes/objects makes it work, but I don't think that's a viable solution.

gabro:
The only caveats are about nesting those annotated objects/classes inside other classes or lambdas, but these are plain top-level objects/classes so it should just work.

Reading through the implementation seems to confirm this...

https://github.com/scala-native/scala-native/blob/895b3673ac1844d02e653f83d108ef5466c5e5d3/nscplugin/src/main/scala/scala/scalanative/nscplugin/NirGenStat.scala#L211-L219

gabro:
@larsrh fixed! It was an issue with the fully qualified name which lacked the final `$`. Also, apparently now the arguments meant for the ScalaCheck runner need to be surrounded in quotes, or they will crash sbt's test-interface parser

gabro:
Any god to pray to run the CI? It seems it didn't like my force push 😅

gabro:
@larsrh one option would be to merge (using the rebase option to keep the history linear) so that my commits end up onto your PR (I've already targeted the right branch).

Maybe that will wake the CI up

PR #756, opened 2021-02-11
Add Cogen.domainOf (again)

nigredo-tori:
See #589. This is a copy of #590, which fell through the cracks at some point.


larsrh:
Thanks!

PR #765, opened 2021-02-27
Update to Scala 2.13.5

ashawley:
Fixes #764 ;-)

### DIFF
--- a/examples/commands-nix/build.sbt
+++ b/examples/commands-nix/build.sbt
@@ -1,6 +1,6 @@
 name := "commands-nix"
 
-scalaVersion := "2.12.13"
+scalaVersion := "2.13.5"
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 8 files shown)


examples/commands-nix/build.sbt
```diff
@@ -1,6 +1,6 @@
 name := "commands-nix"
 
-scalaVersion := "2.12.13"
+scalaVersion := "2.13.5"
 
 libraryDependencies ++= Seq(
   "org.scalacheck" %% "scalacheck" % "1.15.1" % "test",
```
SethTisue, line 6: this doesn't actually have to do with the changes in this PR, but I happened to notice that the 1.15.1 here (and perhaps in the other examples too?) is out-of-date

larsrh (approved):

examples/commands-nix/build.sbt
SethTisue: #766

PR #767, opened 2021-02-28
Fix typos

bwignall:
Found/fixed three typos, two in doc strings (and therefore non-semantic) and one in a `property` string. If the latter breaks anything, happy to back it out.


SethTisue (approved):

larsrh:
Thanks!

PR #769, opened 2021-03-09
Fix Off By One Error In JavaTimeChoose

### LINKED ISSUE #762: Invalid bounds in JavaTimeChoose
Version: 1.15.3

Caused by: org.scalacheck.Gen$Choose$IllegalBoundsError: invalid bounds: low=1, high=0
	at org.scalacheck.Gen$Choose$$anon$8.choose(Gen.scala:475)
	at org.scalacheck.Gen$Choose$$anon$8.choose(Gen.scala:473)
	at org.scalacheck.Gen$Choose$$anon$12.choose(Gen.scala:596)
	at org.scalacheck.Gen$.choose(Gen.scala:637)
	at org.scalacheck.time.JavaTimeChoose$$anon$2.$anonfun$choose$3(JavaTimeChoose.scala:68)
	at org.scalacheck.time.JavaTimeChoose$$anon$2.$anonfun$choose$3$adapted(JavaTimeChoose.scala:55)

Looks like the case where, in #chooseInstant, epochSecond != min.getEpochSecond and epochSecond == max.getEpochSecond == 0.

isomarcte:
The minimum value of nanoseconds for both Duration and Instant should be 0, unless the second value == the minimum bound's second value. It was incorrectly encoded as 1.

Fixes #762

### DIFF
--- a/jvm/src/main/scala/org/scalacheck/time/JavaTimeChoose.scala
+++ b/jvm/src/main/scala/org/scalacheck/time/JavaTimeChoose.scala
@@ -27,7 +27,7 @@ private[scalacheck] trait JavaTimeChoose {
                 if(seconds == minSeconds) {
                   min.getNano
                 } else {
-                  1
+                  0
                 }
               val maxNanos: Int =
                 if (seconds == maxSeconds) {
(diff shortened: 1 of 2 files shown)


isomarcte:
@larsrh @ashawley any thoughts here?

ashawley (changes requested):
Worth adding a test or improving an exiting one to capture the failure?

isomarcte:
@ashawley will do

isomarcte:
@ashawley test added

ashawley (approved):

ashawley:
Thanks for adding tests.

PR #771, opened 2021-03-19
Add Gen.bool

mvalle:
The `Gen` object currently doesn't have a generator for Booleans that's simple and easy to find.

This adds a boolean generator to the `Gen` object called `bool`.


ashawley:
I've thought of this, too, but it's not necessary.  

If it's not obvious, we could add documentation that suggests a boolean generator via:

- `Gen.oneOf(true, false)`
- `Arbitrary.arbitrary[Boolean]`

To that point, the scaladoc for Gen and Arbitrary could use some. Don't believe they have any.  

The boolean cases above would be good code examples to have. ;-)

PR #778, opened 2021-04-03
Initialize StringBuilder to suitable size

martijnhoekstra:


larsrh (approved):

src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -1030,13 +1030,13 @@ object Gen extends GenArities with GenVersionSpecific {
 
   def stringOfN(n: Int, gc: Gen[Char]): Gen[String] =
     gen { (p, seed) =>
-      mkString(n, new StringBuilder, gc, p, seed)
+      mkString(n, new StringBuilder(n), gc, p, seed)
```
ashawley, line 1033: The behavior of `Gen.stringOfN` may allow negatives atm.  Might a negative ''n'' throw an exception?

ashawley (changes requested):

larsrh:
Thanks @martijnhoekstra!

PR #781, opened 2021-04-06
Cleanup of time

ashawley:
These are some follow-up changes to the java.time.* contribution in #734.  They're mostly naming cleanup, but it also includes putting the Cogen tests together in one place, and also adopting a ChooseSpecification for other existing tests as a way to lighten up the huge GenSpecification suite.


ashawley:
I'm going to merge.  I have further small followup changes.

PR #785, opened 2021-04-08
Add Arbitrary[Symbol]

### LINKED ISSUE #243: Arbitrary instance for Symbol
Is there a specific reason that there is no default `Arbitrary` instance for `Symbol`? If not, could you just add something along the lines of 

``` Scala
implicit lazy val arbSymbol: Arbitrary[Symbol] =
  Arbitrary(arbitrary[String].map(Symbol(_)))
```

?

ashawley:
Fixes #243.

Quoted symbol is deprecated in Scala 2.13, but the Symbol class is still available.


ashawley:
There was a reference to a binary compatibility issue, but I'm not sure why MiMa would report that as being the case.

ashawley:
Build passes with the latest MiMa 0.8.1.  When Erik tried this 5 years ago it was MiMa 0.1.9, fwiw.

larsrh (approved):

larsrh:
I think the issue 5 years ago was the trait encoding of Scala pre-2.12 :wink:

PR #786, opened 2021-04-08
Drop the sbt-dotty plugin

### LINKED ISSUE #784: Drop the dotty plugin?
Upgrading to sbt 1.5 in #780 gives a warning about sbt-dotty plugin being deprecated:

    The sbt-dotty plugin is no longer neeeded with sbt >= 1.5, please remove it from your build.

ashawley:
Fixes #784.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,15 +1,10 @@
 sourceDirectory := file("dummy source directory")
 
-val scalaMajorVersion = SettingKey[Int]("scalaMajorVersion")
-
-scalaVersionSettings
-
 val Scala212 = "2.12.13"
 val Scala213 = "2.13.5"
-val DottyOld = "3.0.0-RC1"
-val DottyNew = "3.0.0-RC2"
+val Scala30 = "3.0.0-RC2"
 
-ThisBuild / crossScalaVersions := Seq(DottyOld, DottyNew, Scala212, Scala213)
+ThisBuild / crossScalaVersions := Seq(Scala30, Scala212, Scala213)
 ThisBuild / scalaVersion := (ThisBuild / crossScalaVersions).value.last
 
 ThisBuild / githubWorkflowPublishTargetBranches := Seq()
@@ -18,8 +13,9 @@ val PrimaryOS = "ubuntu-latest"
 ThisBuild / githubWorkflowOSes := Seq(PrimaryOS)
 
 val Java8 = "adopt@1.8"
+val Java11 = "adopt@1.11"
 
-ThisBuild / githubWorkflowJavaVersions := Seq(Java8, "adopt@1.11")
+ThisBuild / githubWorkflowJavaVersions := Seq(Java8, Java11)
 
 // we don't need this since we aren't publishing
 ThisBuild / githubWorkflowArtifactUpload := false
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -1,15 +1,10 @@
...
-scalaVersionSettings
-
 val Scala212 = "2.12.13"
 val Scala213 = "2.13.5"
-val DottyOld = "3.0.0-RC1"
-val DottyNew = "3.0.0-RC2"
+val Scala30 = "3.0.0-RC2"
 
-ThisBuild / crossScalaVersions := Seq(DottyOld, DottyNew, Scala212, Scala213)
+ThisBuild / crossScalaVersions := Seq(Scala30, Scala212, Scala213)
```
larsrh, line 7: We need to decide when we want to do this change, because building for anything less than two Dotty versions is going to be extremely painful for downstream libraries (which is virtually everyone). We might have to wait for the next Dotty RC to hit.

build.sbt
```diff
@@ -177,27 +167,16 @@ lazy val sharedSettings = MimaSettings.settings ++ scalaVersionSettings ++ Seq(
...
-    else Set(
-      "1.14.3",
-      "1.15.1",
-      "1.15.2",
-      "1.15.3"
-    ).map(v => "org.scalacheck" %%% "scalacheck" % v)
+    // TODO: re-enable MiMa for Scala 3 once there is a final version
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((3, _)) => Set()
+      case _            => Set("org.scalacheck" %%% "scalacheck" % "1.15.3")
```
larsrh, line 178: Why did you remove the other versions? I specifically added those in #761 to avoid any bloopers like we had before.

build.sbt
ashawley: Ok, I'm out of touch.  I thought this part of the build was an artifact of switching from 0.27 and I assumed things were converging now that Dotty had reached release candidate status.  What's the need for publishing multiple Dotty versions?

SethTisue: It's not strictly necessary, but it allows people to bump their ScalaCheck version separately from bumping their Scala 3 version. This can help cut down the number of possible causes to be investigated when something goes wrong.

The old Dotty / Scala 3 versions do rapidly become obsolete, but sometimes a particular one has a bug that's a showstopper for a particular project, and sometimes people are a version behind while they wait for long dependency chains to publish.

Summary: publishing for the last two prereleases isn't sufficiently important to do if it involves real difficulty, but if it's merely a matter of leaving both versions in a matrix, then we might as well.

ashawley: Ok, the convenience for version flexibility make sense. Happy to put back.

build.sbt
ashawley: I wasn't sure of the purpose.  We have been bitten by compatibility issues, but I recall those are more often from 1) upstream changes in MiMa, 2) things sneak in with version-specific code, and/or 3) because bootstrapping major Scala/Dotty release candidates or finals means ScalaCheck doesn't have a previous artifact published for MiMa to compare.

build.sbt
```diff
@@ -236,25 +220,28 @@ lazy val js = project.in(file("js"))
   .settings(
     Global / scalaJSStage := FastOptStage,
     libraryDependencies +=
-      ("org.scala-js" %% "scalajs-test-interface" % scalaJSVersion).withDottyCompat(scalaVersion.value)
+      ("org.scala-js" %% "scalajs-test-interface" % scalaJSVersion).cross(CrossVersion.for3Use2_13)
```
larsrh, line 223: Annoying that `scalajs-test-interface` isn't cross-built, but this works well enough I'd say.

larsrh (approved):

PR #787, opened 2021-04-08
Fix stack overflow with Prop.all

### LINKED ISSUE #748: Prop.all is not stack-safe
The current implementation of `Prop.all` is not stack-safe:

https://github.com/typelevel/scalacheck/blob/281d93c21fdde374c9c6304fdd02ca5136bf5819/src/main/scala/org/scalacheck/Prop.scala#L428-L431

The `_ && _` expression creates a long chain of `flatMap` calls, which may blow the stack.

Note that the implementation changed in #531 (cc @non), and the previous implementation was stack-safe (it used the `&&` operation on `Prop.Result` instead of `Prop` itself).

I believe we could write a stack-safe implementation of [`Prop#combine`](https://github.com/typelevel/scalacheck/blob/281d93c21fdde374c9c6304fdd02ca5136bf5819/src/main/scala/org/scalacheck/Prop.scala#L60-L61), which is used by [`Prop#&&`](https://github.com/typelevel/scalacheck/blob/281d93c21fdde374c9c6304fdd02ca5136bf5819/src/main/scala/org/scalacheck/Prop.scala#L122).

ashawley:
Revert the refactoring done to Prop.all and Prop.atLeastOne in #531.

Fixes #748.

### DIFF
--- a/src/main/scala/org/scalacheck/Prop.scala
+++ b/src/main/scala/org/scalacheck/Prop.scala
@@ -427,13 +427,15 @@ object Prop {
 
   /** Combines properties into one, which is true if and only if all the
    *  properties are true */
-  def all(ps: Prop*): Prop =
-    ps.foldLeft(proved)(_ && _)
+  def all(ps: Prop*) =
+    if (ps.isEmpty) proved
+    else Prop(prms => ps.map(p => p(prms)).reduceLeft(_ && _))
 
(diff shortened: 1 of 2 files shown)


larsrh:
Can we add a test for this to make sure it doesn't regress?

ashawley:
Good idea.  There was a test.  I changed it to run once with a million Props.

PR #788, opened 2021-04-08
Hide ambiguous implicit buildableSeq

### LINKED ISSUE #721: Ambiguous implicit values
Seth ran across an issue in the Scala build when upgrading to ScalaCheck 1.15.1 from 1.14.3 in https://github.com/scala/scala/pull/9301.

Seems it fails to compile because of an ambiguous implicit:

```
[error] /home/jenkins/workspace/scala-2.13.x-validate-main@2/test/scalacheck/scala/collection/FloatFormatTest.scala:86:17: ambiguous implicit values:
[error]  both method buildableArrayList in object Buildable of type [T]org.scalacheck.util.Buildable[T,java.util.ArrayList[T]]
[error]  and method buildableSeq in object Buildable of type [T]org.scalacheck.util.Buildable[T,scala.collection.Seq[T]]
[error]  match expected type org.scalacheck.util.Buildable[String,C]
[error]     Gen.sequence(bogoparts).map(_.asScala.mkString)
[error]                 ^
[error] /home/jenkins/workspace/scala-2.13.x-validate-main@2/test/scalacheck/scala/collection/IntegralParseTest.scala:123:33: ambiguous implicit values:
[error]  both method buildableArrayList in object Buildable of type [T]org.scalacheck.util.Buildable[T,java.util.ArrayList[T]]
[error]  and method buildableSeq in object Buildable of type [T]org.scalacheck.util.Buildable[T,scala.collection.Seq[T]]
[error]  match expected type org.scalacheck.util.Buildable[Char,C]
[error]     val sequenced = Gen.sequence(listOfGens)
[error]                                 ^
[error] two errors found
```

ashawley:
Fixes #721.

### DIFF
--- a/src/main/scala/org/scalacheck/util/Buildable.scala
+++ b/src/main/scala/org/scalacheck/util/Buildable.scala
@@ -27,7 +27,7 @@ object Buildable extends BuildableVersionSpecific {
       def builder = new ArrayListBuilder[T]
     }
 
-  implicit def buildableSeq[T]: Buildable[T, Seq[T]] =
+  private[scalacheck] implicit def buildableSeq[T]: Buildable[T, Seq[T]] =
     new Buildable[T, Seq[T]] {
       def builder: mutable.Builder[T, Seq[T]] =
         Seq.newBuilder[T]
--- a/project/MimaSettings.scala
+++ b/project/MimaSettings.scala
@@ -19,6 +19,7 @@ object MimaSettings {
   )
 
   private def removedPrivateMethods = Seq(
+    "org.scalacheck.util.Buildable.buildableSeq"
   )
 
   private def removedPrivateClasses = Seq(
--- a/src/test/scala/scala/StringSpecification.scala
+++ b/src/test/scala/scala/StringSpecification.scala
@@ -1,0 +1,27 @@
+/*-------------------------------------------------------------------------*\
+**  ScalaCheck                                                             **
+**  Copyright (c) 2007-2021 Rickard Nilsson. All rights reserved.          **
+**  http://www.scalacheck.org                                              **
+**                                                                         **
+**  This software is released under the terms of the Revised BSD License.  **
+**  There is NO WARRANTY. See the file LICENSE for the full text.          **
+\*------------------------------------------------------------------------ */
+
+package scala
+
+import org.scalacheck.Gen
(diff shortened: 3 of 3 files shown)


src/main/scala/org/scalacheck/util/Buildable.scala
```diff
@@ -27,7 +27,7 @@ object Buildable extends BuildableVersionSpecific {
       def builder = new ArrayListBuilder[T]
     }
 
-  implicit def buildableSeq[T]: Buildable[T, Seq[T]] =
+  private[scalacheck] implicit def buildableSeq[T]: Buildable[T, Seq[T]] =
```
larsrh, line 30: This is a no-go as far as binary compatibility is concerned. Instead, I propose:

* make the definition non-implicit
* slap a `@deprecated` on it

larsrh (changes requested):

src/main/scala/org/scalacheck/util/Buildable.scala
ashawley: It's an internal definition, so it needs to exist as an implicit.

ashawley: I don't think there's any reason to add `@deprecated` since nobody likely uses it, but I'll do so just to be thorough.

ashawley: Making it package-private is technically binary compatible, it's just not source compatible.  It's an obscure method in an internal object, so I don't think there will be anyone who will even notice the source-breaking change.

larsrh: It's apparently not binary compatible in this instance, since you had to define a MiMa exception. But your argument that it this definition is not likely to be used publicly is persuasive :+1:

larsrh (approved):

src/main/scala/org/scalacheck/util/Buildable.scala
ashawley: It actually it is binary compatible.  Mima is just being thorough.  The emitted byte code is just not binary compatible for Java callers.  ScalaCheck doesn't have any of those.  And if we did, they still likely weren't using it for this implicit.

PR #791, opened 2021-04-10
silence MiMa warning

larsrh:
`MimaPlugin` is enabled automatically.


ashawley (approved):

ashawley:
That seems right.  We don't need to do `enablePlugins(MimaPlugin)`.

I checked out your branch, and tried removing something and I got errors from Mima, so

PR #783, opened 2021-04-07
Add docs for Gen

ashawley:
Fixes #773.

PR #792, opened 2021-04-10
fix CI build script

larsrh:
Turns out those extra quotes didn't do what we thought they would.

### DIFF
--- a/tools/travis-script.sh
+++ b/tools/travis-script.sh
@@ -9,7 +9,7 @@ else
   TESTS=1000
 fi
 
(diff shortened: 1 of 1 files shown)


ashawley (approved):

ashawley:
Good catch, Lars!!!

larsrh:
Random catch, I was looking through some CI log and it complained about invalid arguments :sweat_smile:

ashawley:
That explains what I saw in #783, then?

larsrh:
Possibly. If you rebase that one you should also see the Scaladoc error.

PR #793, opened 2021-04-11
ScalaCheck follows semver since 1.14.x

larsrh:
Set PVP to avoid false negatives, as pointed out by @julienrf.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -85,6 +85,8 @@ def env(name: String): Option[String] =
 
 val isRelease = env("IS_RELEASE").exists(_ == "true")
 
+ThisBuild / versionScheme := Some("pvp")
+
 lazy val sharedSettings = MimaSettings.settings ++ Seq(
 
   name := "scalacheck",


ashawley (approved):

larsrh:
I'm not actually sure whether the change is correct, so I've started asking around.

ashawley:
Is there a policy for MiMa that says "we're binary compatible for downstream Scala libraries but might not always be source compatible with major versions"?  I'm half joking here.

larsrh:
My concern is not so much about our policy, but rather sbt's interpretation of this attribute. The question is: will sbt yell if 1.13.0 gets evicted by 1.15.4? Those are incompatible, but if we slap "semver" on releases from now on, it might do that.

ashawley:
So should we wait to enable this in 1.16.0?

ashawley:
According to <https://github.com/scalacenter/sbt-version-policy>, we need to add the `versionPolicyCheck` task to the build.

ashawley:
Oh, that's an entirely different plugin.  Disregard that.

larsrh:
> So should we wait to enable this in 1.16.0?

We might have to wait until 2.0.0, since we switched bincompat policy in the middle of a major version.

julienrf:
> My concern is not so much about our policy, but rather sbt's interpretation of this attribute. The question is: will sbt yell if 1.13.0 gets evicted by 1.15.4? Those are incompatible, but if we slap "semver" on releases from now on, it might do that.

If you declare your `versionScheme` to be `semver-spec`, no sbt will not fail if 1.13.0 is evicted by 1.15.4, since the later is assumed to be backward binary compatible with the former, according to `semver-spec`.

PR #794, opened 2021-04-16
Only run bench once in the build

ashawley:
Pare down the build.  It's not necessary to run across Scala versions.  

Running against the latest Scala is what we did in the old Travis build, see #609.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


larsrh (approved):

.github/workflows/ci.yml
```diff
@@ -162,5 +162,5 @@ jobs:
             ~/Library/Caches/Coursier/v1
           key: ${{ runner.os }}-sbt-cache-v2-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
 
-      - name: Run dubious benchmark suite
+      - name: Build benchmark suite
```
SethTisue, line 165: 😆

PR #797, opened 2021-04-19
Drop doc setting in native

ashawley:
Saw this in the Native build and not sure what it does.  

```scala
Compile / doc := (jvm / Compile / doc).value
```

Maybe it used to do something, but is no longer necessary?


larsrh (approved):

larsrh:
Probably a leftover from when we had to disable Dottydoc.

SethTisue:
(Or it might have been an effort to save time by not generating the same doc twice?)

PR #814, opened 2021-05-13
Adds Discord badge

paualarco:
Making the Discord support channel more visible by adding a badge with a badge that links to #scalacheck and that never expires :)

This is how it looks like:

Screen Shot 2021-05-14 at 00 09 57


larsrh:
Thanks!

PR #825, opened 2021-06-24
Fix Arbitrary[Year] instance max value

dantb:
Caused a very narrow range of inputs in some tests I was writing 😄


larsrh:
Good catch!

PR #830, opened 2021-07-29
Support `java.time` arbitraries on Scala.js

armanbilge:
We can also use these instances on Scala.js via [scala-java-time](https://github.com/cquiroz/scala-java-time). This would be helpful for https://github.com/http4s/http4s/pull/4938. 

I was surprised that most of the tests for this project (not just specific to `java.time`) are only run on the JVM, is there a particular reason for this? It would be great to run more of these on more platforms.

Thanks!


ashawley:
When time support was added in #734, it was known that Scala.js could be supported.  The issue was deferred was all.

armanbilge:
Absolutely! Well if you are open to it now it would be really helpful :)

armanbilge:
I think the CI flaked :(
https://github.com/typelevel/scalacheck/runs/3196039194?check_suite_focus=true#step:6:31

ashawley:
Yeah, it does that on occasion.  Re-starting...

armanbilge:
I think it happened again 😅

armanbilge:
Sadly, I have to close this :( @mpilquist pointed out to me that adding a dependency to scala-java-time here creates a circular dependency between it and scalacheck. The right thing to do is to provides these arbitraries in scala-java-time itself.

armanbilge:
Revisiting this, I realized that actually we don't need scala-java-time here. So there is no circular dependency!

jvm-js/main/scala/org/scalacheck/time/JavaTimeArbitrary.scala
```diff
@@ -25,14 +25,11 @@ private[scalacheck] trait JavaTimeArbitrary {
...
-
-  private final lazy val maxJavaDuration: Duration =
-    Duration.ofSeconds(Long.MaxValue, 999999999L)
-
-  implicit final lazy val arbJavaDuration: Arbitrary[Duration] =
+  implicit final lazy val arbJavaDuration: Arbitrary[Duration] = {
+    val minJavaDuration = Duration.ofSeconds(Long.MinValue)
+    val maxJavaDuration = Duration.ofSeconds(Long.MaxValue, 999999999L)
     Arbitrary(Gen.choose(minJavaDuration, maxJavaDuration))
+  }
```
armanbilge, line 32: The Scala.js linker/optimizer was unable to elide the `Duration` reference for some reason. This fixed it.

armanbilge:
Hmm, nothing changed? Just my own understanding. In my initial attempt I was adding an unneeded dependency.

ashawley:
Ok, I was wondering if scala-java-time or similar was recently merged with Scala.js internals.  If so, was curious if it required a minimum version of Scala.js for these time bindings.

armanbilge:
So FTR my confusion was about compile time vs linking/runtime. Since we are not testing these arbitraries, we don't need `java.time` implementations for Scala.js, only the types signatures for compilation. These are provided by the JDK itself.

ashawley:
Ok, thanks for the clarification.

ashawley:
Do you know if Scala native is similarly able to support Java time API in this way?  Should we do that here, as well, while we're at it?

armanbilge:
Yes, I think so. My understanding has been that SJS and native have a similar model wrt compile vs link/run-time. Let me do that here, actually it should simplify this PR!

In any case, it seems the `java.time` implementations for native are provided by a different library, which would not create a circular dependency with scalacheck.
https://github.com/ekrich/sjavatime

ashawley:
Great, thanks for looking into it.

armanbilge:
Cool, that seems to have worked :) good idea!

armanbilge:
@rossabaker while you're here this one is good for http4s 😉

rossabaker (approved):
I barely understand scala.JS and native not at all, but it's well argued and got a green CI run.

PR #835, opened 2021-09-14
Fix Arbitrary[Char] generates 0xFFFE (not a character in Unicode)

kadzuya:
## Problem 

Valid Char ranges are [0x0000, 0xD7FF] and [0xE000, 0xFFFD].
But Arbitrary[Char] generates 0xFFFE since  #653.

```scala
scala> (63486 + 2048).toHexString
val res0: String = fffe
```

## Test

I tested with 11e23c7 (only add test for Arbitrary[Char]), it failed.
(This test with default parameters will almost always succeed...)
```
$ sbt 'jvm / Test / console'

scala> import org.scalacheck.Test.Parameters
import org.scalacheck.Test.Parameters

scala> org.scalacheck.ArbitrarySpecification.check(Parameters.default.withMinSuccessfulTests(100000))
+ Arbitrary.arbOption coverage: OK, proved property.
failing seed for Arbitrary.arbChar is a0y7J3eh7ctLsQdJDj0uwLsuqfXp9a0rswkVdkFKxOK=
! Arbitrary.arbChar: Falsified after 1093 passed tests.
> ARG_0: ￾
+ Arbitrary.arbString: OK, passed 100000 tests.
+ Arbitrary.arbSymbol: OK, passed 100000 tests.
+ Arbitrary.arbitrary[Recur].passes: OK, passed 100000 tests.
+ Arbitrary.arbitrary[Recur].throws[StackOverflowError]: OK, proved propert
  y.
+ Arbitrary.arbEnum: OK, proved property.
```

with 902121e (this PR), it succeeded.

```
scala> org.scalacheck.ArbitrarySpecification.check(Parameters.default.withMinSuccessfulTests(100000))
+ Arbitrary.arbOption coverage: OK, proved property.
+ Arbitrary.arbChar: OK, passed 100000 tests.
+ Arbitrary.arbString: OK, passed 100000 tests.
+ Arbitrary.arbSymbol: OK, passed 100000 tests.
+ Arbitrary.arbitrary[Recur].passes: OK, passed 100000 tests.
+ Arbitrary.arbitrary[Recur].throws[StackOverflowError]: OK, proved propert
```

### DIFF
--- a/src/main/scala/org/scalacheck/Arbitrary.scala
+++ b/src/main/scala/org/scalacheck/Arbitrary.scala
@@ -152,8 +152,8 @@ private[scalacheck] sealed trait ArbitraryLowPriority {
   implicit lazy val arbChar: Arbitrary[Char] = Arbitrary {
     // valid ranges are [0x0000, 0xD7FF] and [0xE000, 0xFFFD].
     //
-    // ((0xFFFD + 1) - 0xE000) + ((0xD7FF + 1) - 0x0000)
-    choose(0, 63486).map { i =>
+    // ((0xFFFD + 1) - 0xE000) + ((0xD7FF + 1) - 0x0000) - 1
+    choose(0, 63485).map { i =>
       if (i <= 0xD7FF) i.toChar
       else (i + 2048).toChar
     }
--- a/jvm/src/test/scala/org/scalacheck/ArbitrarySpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/ArbitrarySpecification.scala
@@ -9,9 +9,10 @@
 
 package org.scalacheck
 
+import java.util.concurrent.TimeUnit
+
 import Prop._
 import Arbitrary._
-import java.util.concurrent.TimeUnit
 
 object ArbitrarySpecification extends Properties("Arbitrary") {
   val genOptionUnits =
--- a/jvm/src/test/scala/org/scalacheck/ArbitrarySpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/ArbitrarySpecification.scala
@@ -23,6 +23,11 @@ object ArbitrarySpecification extends Properties("Arbitrary") {
   property("arbOption coverage") =
     exists(genOptionUnits) { case (a, b) => a.isDefined != b.isDefined }
 
+  property("arbChar") =
+    Prop.forAll { (c: Char) =>
+      0x0000 <= c && c <= 0xD7FF || 0xE000 <= c && c <= 0xFFFD
+    }
+
   property("arbString") =
     Prop.forAll { (s: String) =>
       s ne new String(s)


ashawley:
> (This test with default parameters will almost always succeed...)
>     scala> org.scalacheck.ArbitrarySpecification.check(Parameters.default.withMinSuccessfulTests(100000))
>    ...

FYI this can also be run from the sbt console with:

    sbt> jvm/testOnly org.scalacheck.ArbitrarySpecification -- -s 100000

src/main/scala/org/scalacheck/Arbitrary.scala
```diff
@@ -152,8 +152,8 @@ private[scalacheck] sealed trait ArbitraryLowPriority {
   implicit lazy val arbChar: Arbitrary[Char] = Arbitrary {
     // valid ranges are [0x0000, 0xD7FF] and [0xE000, 0xFFFD].
     //
-    // ((0xFFFD + 1) - 0xE000) + ((0xD7FF + 1) - 0x0000)
-    choose(0, 63486).map { i =>
+    // ((0xFFFD + 1) - 0xE000) + ((0xD7FF + 1) - 0x0000) - 1
+    choose(0, 63485).map { i =>
```
ashawley, line 156: So it was just a miscalculation of the range that was missed, or was `Gen.choose` not inclusive at some point?

ashawley: I think I just need to spend time to grok the math.

ashawley:
I like the new test, even if it doesn't catch the regression with default settings.  If we had had the obvious test, it probably would have eventually failed at least once over the last 2 years.

src/main/scala/org/scalacheck/Arbitrary.scala
ashawley: Oh, so zero-based iteration from 0 to i needs to be exclusive upper bound, but `Gen.choose` is inclusive bounds.  The range is actually 63486 values, but we're computing from zero.

ashawley: By example, the ranges of [0, 2] and [6, 8] are actually a total of 6 values:

(_2_ - _0_ + _1_) + (_8_ - _6_ + _1_) = _3_ + _3_ = _6_

However, zero-based iteration through the range of values should be upper bound exclusive, [0, 6).

```
scala> (0 to 5).map { i =>                                                      
     |   if (i <= 2) i
     |   else i + 3
     | }
     |
val res0: IndexedSeq[Int] = Vector(0, 1, 2, 6, 7, 8)
```

ashawley (approved):

ashawley:
Thanks for the contribution.

src/main/scala/org/scalacheck/Arbitrary.scala
kadzuya: Yes, that's the cause.
I wrote almost the same code during development. 😄

rossabaker (approved):

PR #845, opened 2021-10-15
Vary the length of Gen.listOfN

### LINKED ISSUE #844: Behavior of Gen.listOfN behaves different to what I was assuming based on the documentation
I have a Generator for a  list of elements. To limit the number of elements in that list, I used `Gen.listOfN` instead of `Gen.listOf`.

[The Scaladoc](https://github.com/typelevel/scalacheck/blob/main/src/main/scala/org/scalacheck/Gen.scala#L1050) says _"Generates a list with at most the given number of elements"_, so I was assuming that the generated lists will have different lengths, but never more than the provided value for the argument `n`. 
But instead what I see is that the lists always seem to be of length `n`.

I would assume that this is the intended behavior and not a bug (based on a quick look at `buildableOfN`, it seems to be intended to aim for `n` elements in the result), so probably I was fooled by the docs.

Theoretically I could go with `Gen.list(...).map(_.take(n))`, but that wouldn't be efficient because the first generator would generate a lot of values that are discarded by `take`.

Or something like:

```scala
def listGen[T](n: Int)(implicit gen: Gen[T]): Gen[List[T]] = {
  for {
    length <- Gen.choose(0, n)
    ts <- Gen.listOfN(length, gen)
  } yield ts
}
```
which feels primitive enough to make me expect it in scalacheck directly.

Am I missing something here? Is there a different way to write an efficient generator for a list that produces lists of different lengths with a given maximum length?

ashawley:
Fixes #844.


joroKr21:
I will just weigh in my opinion as a user - I think changing the behaviour in this way has the potential to break too many tests. Also the current behaviour is consistent with the name `listOfN`. For these reasons I think it would be better to just change the documentation of `listOfN` and add new methods with "at most" semantics. If it's possible, calling `buildableOfN` for `Set` and `Map` could be deprecated.

ashawley:
I agree.  I raised the same concern of breaking tests in #844.  The chance of generating lists less than _N_ with `listOfN` is possible but usually pretty remote and actually is only possible on what your combined generator is or what the test is.

satorg:
@ashawley just to clarify with you: do we want to continue on this PR or do you think it can be closed without merging?
Thank you for your efforts in any case!

ashawley:
Yes, this can be closed as unmerged. 

It was exploring fixing #844, but fixing it would have been fraught for users.

PR #849, opened 2021-10-22
Add continuously growing collection Generators

mdedetrich:
This PR adds generators which allow you to generate a collection/list that continuously grows until a condition/predicate is fullfilled (with the input of the condition being the current state of the collection being generated).

I am creating this PR because I currently have a usecase where I need to create a collection that is of specific size (in bytes, not number of elements) due to testing multipart uploads and having to make sure that generated test data is of least a specific chunk size but I suppose the problem is generic enough that it can apply to other issues.

Due to `Gen` being a `sealed trait` and certain methods inside `Gen` being private its not that easy to extend `Gen` with your own generator methods.

The naming of the methods can probably be done better, also not sure if I need to add tests for these and if so what I should be testing for/what they should look like. The `failureHint`'s approach is also an open question due to the fact that its obviously not possible to figure out the size of the resulting collection.


mdedetrich:
@armanbilge Would it be possible to look into this?

satorg:
@mdedetrich but is it really necessary to extend `Gen` for that? There's a _public_ method `infiniteLazyList` (or `infitineStream`) in `Gen` that (I think) can be used to achieve pretty much the same functionality, e.g.:
```scala
  def buildableOfCondGen[C, T](fillCondition: C => Boolean, g: Gen[T])(
      implicit
      evb: Buildable[T, C],
      evt: C => Iterable[T]
  ): Gen[C] =
    Gen.infiniteLazyList(g).map { ll =>
      val it = ll.iterator
      val bldr = evb.builder
      while (!fillCondition(bldr.result()))
        bldr += it.next()
      bldr.result() // sub-optimal: is called twice for the same result, can be improved!
    }

  val vgen: Gen[Vector[Int]] = buildableOfCondGen[Vector[Int], Int](_.size == 123, Gen.chooseNum(456, 789))

  Prop.forAll(vgen) { v =>
    v.size == 123
  }.check()
```
And perhaps you even don't have to worry about `failureHint` with it.

mdedetrich:
@satorg

Perfect thanks, I didn't realize there was a flexible `infiniteLazyList` that can theoretically be used to do the same. Will look into this and report back!

satorg:
Hi @mdedetrich , sorry for the long ping. Just to clarify with you: would you like to continue on this PR or do you think it can be closed? Thank you for your effort in any case!

Issue #574, opened 2019-09-27
documentation page has bad links


oscar-stripe:
https://www.scalacheck.org/documentation.html

see for instance "online":

https://www.scalacheck.org/files/scalacheck_2.11-1.14.1-api/index.html

erik-stripe:
Thanks for reporting this.

@rickynils do you know how those files get populated? Do I need to do something during the release? Should I try to host the Scaladocs somewhere else and you can just link to them?

non:
Thanks for reporting this.

@rickynils do you know how those files get populated? Do I need to do something during the release? Should I try to host the Scaladocs somewhere else and you can just link to them?

non:
This ties into a more general problem: we still need @rickynils to update the static site.

It's possible we should move the site to something hosted in Github, so @ashawley and I can manage it. What do you all think?

deanwampler-domino:
I just noticed this same bug, 1.5 years later.

ddl-deanwampler:
And again... Let me know if I can help move it to a GitHub site.

rossabaker:
I could flip the switch to enable publishing to ghpages, but we'd need the domain owner's help to have scalacheck.org point at it.

rickynils:
@rossabaker I can do that, just tell me where I should point the domain

rossabaker:
Thanks, @rickynils.  It turns out it's not as easy as flipping the switch: I didn't find source here to point to.  I fixed that in #850.  If you've been managing this with a static site generator and can link to the source, we could use that instead.

Once that's merged, I can enable GitHub pages, and then we should just need [the apex domain and www subdomain variant](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site#configuring-a-subdomain) set up in your DNS.

rickynils:
@rossabaker The existing site is generated by using a combination of [Hakyll](https://jaspervdj.be/hakyll/) and [Nix](https://nixos.org/manual/nix/stable/). I can make the source available, but I wonder if it wouldn't be better for the current maintainers to start over with something they feel comfortable to maintain?

rossabaker:
I'm comfortable with Nix, but I'm barely a maintainer here.  I'm happy with #850 as an interim solution to give the regular maintainers time to discuss what they're comfortable with.

ashawley:
[...]

PR #850, opened 2021-11-01
Crawl the existing website to host on gh-pages

rossabaker:
To get scalacheck.org moved over to GitHub Pages, we need something source for it.  I couldn't find source for the existing site, so I crawled it.  We can think about something like Laika or Jekyll later if we like, but I wanted to bootstrap something while we have @rickynils' attention.

* Did not include `files/`, which we should be able to link from elsewhere.  This will break more until we do!
* Removed fonts to avoid copyright headaches.  A `<link>` is better than `@import`, but `@import` is DRY.
* Added the necessary CNAME to point scalacheck.org at this


rossabaker:
We got the sources in #574, which is a better solution.

PR #852, opened 2021-11-04
Set up gh-pages with Rickard's Hakyll source

rossabaker:
Another riff on #574.


rossabaker:
1. Does not work on Darwin.  I think I can fix that.  If I can't, I can grab my Linux laptop.
2. I have to build ghc from source to build the site.  Needs something like Cachix if this is going to be our solution.
3. I have to run `nix-shell -p nixUnstable` because the rest of my world can't move to Nix 2.4 yet.

rossabaker:
djspiewak/sbt-github-actions#87 would take care of the wasteful CI.  In the interim, I'll use the `[ci skip]` tag.

rossabaker:
The output from this Nix build is now published to https://typelevel.org/scalacheck/ via the execrable push-site.sh.  When Rickard does the DNS work, we can set up the custom domain and get it back on scalacheck.org.

rossabaker:
I switched to a CDN for fonts to:
1. Avoid figuring out copyright attribution
2. Have a change so we know it's being rendered by GH pages

PR #860, opened 2021-12-07
Set up a proper cross build

vasilmkd:
Fixes #625.

Edited:

This snapshot seems to be working fine on all platforms:

```scala
"io.vasilev" %%% "scalacheck" % "1.15-6862df0"
```

Help for reviewers, list of changes:

- Created a `core` directory where all the code is found (similar to other Typelevel libraries).
- Set up a proper cross build in `core` (one project with three platforms and platform specific managed sources instead of three projects with unmanaged sources)
- `jvm` became the `jvm` cross build under `core/jvm`
- `js` became the `js` cross build under `core/js`
- `native` became the `native` cross build under `core/native`
- `src` became the shared source code for all platforms under `core/shared/src`
- I changed the `scala-2.13-` and `scala-2.13+` unmanaged directories to use the "official" `scala-2.12`, `scala-2.13` and `scala-3` Scala versions specific directories supported by the crossproject plugins (the sources under `scala-3` can be seen as new files in the diff)
- Removed as much cruft as possible from the `build.sbt` file
- Removed the `release.sh` script because `sbt-spiewak-sonatype` contains `sbt-sonatype` which can be used for releases (already used it to publish a snapshot and tested the release)
- Tweaked things in `tools/travis-script.sh` to match the new subprojects (`coreJVM`, `coreJS`, `coreNative`)

There are build warnings which I plan to address in a follow-up PR, to keep this diff as small as possible.

PR #868, opened 2022-01-05
support scala-native Scala 3 build

xuwei-k:


build.sbt
```diff
@@ -253,10 +253,10 @@ lazy val jvm = project.in(file("jvm"))
 lazy val native = project.in(file("native"))
   .settings(sharedSettings: _*)
   .settings(
-    scalaVersion := Scala212,
-    crossScalaVersions := Seq(Scala212, Scala213),
     // TODO: re-enable MiMa for native once published
     mimaPreviousArtifacts := Set(),
+    publish / skip := CrossVersion.partialVersion(scalaVersion.value) == Some((3, 0)),
+    Compile / doc / sources := Nil, // TODO
```
xuwei-k, line 259: - https://github.com/scala-native/scala-native/issues/2503

build.sbt
```diff
@@ -32,7 +32,7 @@ ThisBuild / githubWorkflowBuildMatrixInclusions ++=
   }
 
 ThisBuild / githubWorkflowBuildMatrixInclusions ++=
-  crossScalaVersions.value.filter(_.startsWith("2.")) map { scala =>
+  crossScalaVersions.value.filterNot(_ == Scala30) map { scala =>
```
xuwei-k, line 35: - https://repo1.maven.org/maven2/org/scala-native/nscplugin_3.1.0/
- https://scala-native.readthedocs.io/en/latest/changelog/0.4.3-RC1.html

> The minimal version of Scala 3 supported by Scala Native is 3.1.0, due to fatal blockers in Scala 3.0.x.

vasilmkd:
@xuwei-k Is it maybe easier to base this PR on top of #860, which unfortunately wasn't looked at by any other maintainer.

build.sbt
vasilmkd: If it were up to me, I would release Scala 3 native with a minor version increase (1.16.0), which would only be published for Scala 3.1.

kpodsiad:
What is the current status of this PR? Having scala-native support will be awesome.

rossabaker:
Is this PR still necessary if we merge https://github.com/typelevel/scalacheck/pull/875?

armanbilge:
This PR requires Scala 3.1.

SethTisue:
Can we use Scala 3.0.x for all of the CI matrix entries _except_ for the Scala 3 + Scala Native combination, which would use 3.1.x?

SethTisue:
Needs rebase.

armanbilge:
Maybe, these sorts of hacks seem to cause problems see e.g. https://github.com/portable-scala/sbt-crossproject/issues/130.

After all this build engineering I'm not really into build hacks ... if people really want this now, I think we should all bump to Scala 3.1. It's the first Scala 3 supported on all platforms anyway

armanbilge:
@vasilmkd had an idea that we bump the minor version and go to Scala 3.1. _If_ it becomes a problem, we can continue "maintaining" the other branch ... but things have been moving slow here anyway, not sure what critical maintenance would show up on that branch.

vasilmkd:
I would still advise on bumping the minor version in the release that ships on 3.1, because it signals a forward incompatible change.

armanbilge:
Even if we invoke the `-Yscala-release` flag?

vasilmkd:
I don't trust it without an experiment.

SethTisue:
#879 is merged (moving us to 3.1), so I guess this should be rebased now?

armanbilge:
https://github.com/typelevel/scalacheck/pull/879 actually supersedes this one, so we can close.

PR #874, opened 2022-02-27
Add pick generator specialized for indexed sequences

lambdani:
Hi! Would there be interest in a pick generator specialized for IndexedSeqs? When choosing k elements from a sequence with n elements, the idea is to choose an element in the inclusive range [0,n-1], then another one in [0,n-2]... up to [0,n-k]. Then these indices must be translated to the whole range [0,n-1] while avoiding repetitions. For this, one can use a modified version of an order statistic tree that selects the i-th non negative integer not present in the tree.

This should pick k elements in O(k log k) time, using O(k) extra space for the tree. Additionally, the elements should be permuted in random order.

The names are horrible but I couldn't come up with better ones. Any help with that would be appreciated if you think it's worth to add this generator to Scalacheck. What do you think?

### DIFF
--- a/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -260,6 +260,30 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
     }
   }
 
+  property("indexedPick") = forAll { (vec: Vector[Int]) =>
+    forAll(choose(-1, 2 * vec.length)) { n =>
+      Try(indexedPick(n, vec)) match {
+        case Success(g) =>
+          forAll(g) { m => m.length == n && m.forall(vec.contains) }
+        case Failure(_) =>
+          Prop(n < 0 || n > vec.length)
+      }
+    }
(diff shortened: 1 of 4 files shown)


satorg:
@lambdani just to clarify: is the suggested behavior different from this one?
```scala
def shuffledPick[T](n: Int, seq: IndexedSeq[T]): Gen[collection.Seq[T]] = {
  Arbitrary.arbitrary[Long].flatMap { seed =>
    val shuffledSeq = new scala.util.Random(seed).shuffle(seq)
    Gen.pick(n, shuffledSeq)
  }
}
```
I mean, I realize it works in a different way, but not sure if there's a difference in the results they both will be providing.

lambdani:
No, it should have the same behavior. The only difference should be asymptotic efficiency (O(k log n) vs O(n)). But I don't know if it's faster in practice for enough use cases, and even then if it's worth the added complexity. I could try to write some benchmarks, but it's OK if you think it's not worth it :-).

rossabaker (commented):
Thanks!  A quick benchmark would be interesting to see approximately what size collection is the break-even point.

PR #879, opened 2022-03-18
Onwards and upwards

armanbilge:
Bumps Scala 3 to 3.1.x and Scalacheck to 1.16.x. Closes https://github.com/typelevel/scalacheck/issues/878. Supersedes https://github.com/typelevel/scalacheck/pull/868.

PR #883, opened 2022-03-23
Don't cross-build Native for Scala 3.0

armanbilge:
I think this should fix publishing. So would this one:
- https://github.com/typelevel/scalacheck/pull/879


rossabaker:
Nobody objected to #879, and that's cleaner than this.  I approved that instead.

Issue #886, opened 2022-04-04
Release v1.16.0


armanbilge:
To unblock https://github.com/scalameta/munit/pull/477 and Discipline Native/3 which should get us to Cats Native/3.

rossabaker:
I will cut this tomorrow if nobody tells me to wait.

rossabaker:
Sorry, another bikeshed: is the plugin prepared for tags that don't begin with `v`?  Alternatively, is there any reason we shouldn't switch to the common Typelevel convention now?

armanbilge:
Ah, good question. All tags from this point should start with `v`. MiMa needed some manual intervention to recognize the old artifacts but that's for the better because PVP and stuff.
https://github.com/typelevel/scalacheck/blob/0ac8005753ab98b6494fd631502201b97a103638/build.sbt#L56-L57

rossabaker:
Tagged, but I won't be awake to see the outcome.

rossabaker:
Oh, we should also update the CHANGELOG.  I'll try to remember to do that if it hasn't appeared overnight.

PR #888, opened 2022-04-08
Update Changelog for v1.16.0

armanbilge:
Closes https://github.com/typelevel/scalacheck/issues/886.

I guess I should have PRed the release notes directly here instead of opening an issue, but anyway.


rossabaker:
Rerunning CI.  I suspect this was pushed before Maven Central synced.

PR #889, opened 2022-04-08
Fix the cross-platform time arbitraries, again

armanbilge:
Re-does:
- https://github.com/typelevel/scalacheck/pull/830
- https://github.com/typelevel/scalacheck/pull/882

Between the sbt-typelevel migration, the botched merge, and my blind fury to publish for Native on Scala 3, this slipped through the cracks again. Oh well.


rossabaker (approved):
What are the ramifications of this?  This needs a patch release?

armanbilge:
It's not particularly critical. I was really hoping to get the [`QueryParamCodecSuite`](https://github.com/http4s/http4s/blob/171e581398c68cc1e69144271f6083944dc71335/tests/jvm/src/test/scala/org/http4s/QueryParamCodecSuite.scala) in http4s running cross-platform.

I'm not sure what the versioning rules are for scalacheck. E.g. in Cats(-Effect) new features are only added in minor bumps and patches are only for bugs.

rossabaker:
You're right according to strict semantic versioning, even if it feels patchy.  If you're not begging for a release, we could bundle it with anything else that gets reviewed in the next couple weeks as 1.17.

rossabaker:
Nobody objected.  We've got a motivating change coming through.  Let's ship it and call it 1.17.

PR #894, opened 2022-06-03
[Commands] Fix MatchError when using multiple workers

### LINKED ISSUE #893: MatchError when running Commands property with multiple workers
Encountered this on Scala 2.12 with scalacheck 1.16.0.  For some reason I can't determine, it didn't occur on scalacheck 1.14.3 despite having the same code in Commands.scala

```
val prop = commands.property()
org.scalatest.Test.check(prop)(params => params.withWorkers(2))
```

```
  > Exception: scala.MatchError: (<your state>,None) (of class scala.Tuple2)
  org.scalacheck.commands.Commands.$anonfun$property$3(Commands.scala:222)
```

yannkaiser-asana:
Fixes #893

`for (Some(v) <- Map(...).values) yield v` would throw MatchError as soon as it tries to process a value that doesn't conform to Some(v).

The tests didn't detect this because `property("...") = rhs` was re-evaluating `rhs` each time the property is evaluated, which means that Commands.property() only ran one evaluation at a time (so `suts` would always be empty when we run those for comprehensions).

### DIFF
--- a/core/jvm/src/test/scala/org/scalacheck/commands/CommandsSpecification.scala
+++ b/core/jvm/src/test/scala/org/scalacheck/commands/CommandsSpecification.scala
@@ -13,7 +13,15 @@ import org.scalacheck._
 
 object CommandsSpecification extends Properties("Commands") {
 
-  property("commands") = TestCommands.property(threadCount = 4)
+  def prop = TestCommands.property(threadCount = 4)
+  property("commands") = prop
+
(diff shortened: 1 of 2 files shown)


rossabaker (approved):
Thanks!  This looks like good cause for a patch release.

yannkaiser-asana:
Ready to merge whenever, though it looks like you might need to manually trigger the tests for this PR.

PR #910, opened 2022-07-28
fix wrong match in shrinker (fixes #909)

### LINKED ISSUE #909: Hang while running test suite for `scalatestplus-scalacheck` (possible regression)
Running the test suite for [scalatestplus-scalacheck](https://github.com/scalatest/scalatestplus-scalacheck) appears to hang indefinitely (while shrinking?) when built against the latest commit on the scalacheck main branch (7749c38313f9deb79368d72f592f7f1e9383620d).

The same test suite passes within about 5 seconds if built against the previous merge commit (dc93d7954264e4135be39d699c6130f3160d3af6), so it seems related to #838.

This issue was noticed while updating projects in the Scala 3 community build (https://github.com/lampepfl/dotty/pull/15728#issuecomment-1196246968) however it is also reproducible on Scala 2.13.

### Reproduction Steps

1. `publishLocal` scalacheck @ 7749c38313f9deb79368d72f592f7f1e9383620d
2. clone https://github.com/scalatest/scalatestplus-scalacheck/commit/32e7dc67a9490147e6a04be9125c3ef0940d6819
3. update `build.sbt` of scalatestplus-scalacheck to use the locally published scalacheck:
```diff
-    "org.scalacheck" %%% "scalacheck" % "1.16.0",
+    "org.scalacheck" %%% "scalacheck" % "1.17-7749c38-SNAPSHOT",
```
4. run `sbt scalatestPlusScalaCheckJVM/test`

AurelienRichez:
This fixes the problem introduced by #838 reported in #909. 

The collect I introduced in my previous PR matched on nothing instead of saving the original exception type. As a consequence the the shrink process reads the whole stream provided by the shrinker. I imagine on a shrink with a very long stream (or infinite, but not sure if that exists in the wild) it would just hang.

cc @armanbilge

### DIFF
--- a/core/shared/src/main/scala/org/scalacheck/Prop.scala
+++ b/core/shared/src/main/scala/org/scalacheck/Prop.scala
@@ -773,7 +773,7 @@ object Prop {
     /*
      * Returns the first failed result in Left or success in Right.
      */
-    def getFirstFailure(xs: Stream[T], exceptionFilter: Option[Class[_]]): Either[(T,Result),(T,Result)] = {
+    def getFirstFailure(xs: Stream[T], exceptionFilter: Option[Class[_ <: Throwable]]): Either[(T,Result),(T,Result)] = {
       assert(!xs.isEmpty, "Stream cannot be empty")
       val results = xs.map(x => (x, result(x)))
       results.dropWhile {
@@ -788,7 +788,7 @@ object Prop {
     def shrinker(x: T, r: Result, shrinks: Int, orig: T): Result = {
       val xs = shrink(x)
       val res = r.addArg(Arg(labels,x,shrinks,orig,pp(x),pp(orig)))
-      val originalException = Some(r.status).collect { case NonFatal(e) => e.getClass() }
+      val originalException = Some(r.status).collect { case Prop.Exception(e) => e.getClass }
       if(xs.isEmpty) res else getFirstFailure(xs, originalException) match {
         case Right((x2,r2)) => res
         case Left((x2,r2)) => shrinker(x2, replOrig(r,r2), shrinks+1, orig)
--- a/core/shared/src/test/scala/org/scalacheck/PropSpecification.scala
+++ b/core/shared/src/test/scala/org/scalacheck/PropSpecification.scala
@@ -269,6 +269,30 @@ object PropSpecification extends Properties("Prop") {
     })
   }
 
+  property("shrinking process always take the first value returned by the Shrink[T]") = {
+    // this shrinker split a string in 2
+    implicit val stringShrink: Shrink[String] = Shrink[String] {
+      def split(s: String): Stream[String] = {
+        if (s.length == 1) Stream.empty
+        else {
+          s.take(s.length / 2) #:: s.drop(s.length / 2)  #:: Stream.empty
+        }
+      }
+      split
+    }
+
+    // shrinked value will be "1234", then shrinked to "12" because it also fails, and finally "1"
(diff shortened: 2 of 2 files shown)


core/shared/src/test/scala/org/scalacheck/PropSpecification.scala
```diff
@@ -269,6 +269,30 @@ object PropSpecification extends Properties("Prop") {
     })
   }
 
+  property("shrinking process always take the first value returned by the Shrink[T]") = {
```
AurelienRichez, line 272: This test reproduces the issue. 

With the Shrink provided, normally the shrinking process will go like `12345678` -> `1234` -> `12` -> `1`.

With my previous PR, it checks  something like`12345678` -> `1234` -> `5678`, gets to the end of the stream provided by the custom Shrink and then finally returns `12345678`.  So it's not really the test hanging, but it shows a change in the way the value is shrinked. 

Modifying the stream to be infinite will create a test that never ends. I think that even with finite stream you can see some symptoms as every test will run through the entire stream provided by Shrink instead of choosing the first value.

core/shared/src/main/scala/org/scalacheck/Prop.scala
```diff
@@ -788,7 +788,7 @@ object Prop {
     def shrinker(x: T, r: Result, shrinks: Int, orig: T): Result = {
       val xs = shrink(x)
       val res = r.addArg(Arg(labels,x,shrinks,orig,pp(x),pp(orig)))
-      val originalException = Some(r.status).collect { case NonFatal(e) => e.getClass() }
+      val originalException = Some(r.status).collect { case Exception(e) => e.getClass }
```
armanbilge, line 791: I stared at this yesterday and today and pulled up the PR on my local and added `println`s before I finally understood why such an innocent looking change makes a huge difference 😂

This is very confusing. The `Exception` here is _NOT_ in the `Throwable` hierarchy. That's why `NonFatal(_)` was effectively useless.
https://github.com/typelevel/scalacheck/blob/4867326242450a54f66a5ecbccc2a59dc4630a64/core/shared/src/main/scala/org/scalacheck/Prop.scala#L300

Oddly enough I don't think scalac warns for this?

armanbilge (approved):
Thank you so much for fixing, and your patience on the original PR. LGTM!

core/shared/src/main/scala/org/scalacheck/Prop.scala
armanbilge: Perhaps if you write it as `Prop.Exception` that will help with readability? Perhaps the reviewer in https://github.com/typelevel/scalacheck/pull/838#discussion_r913161658 was confused as well 😝 

Also I now realize you explained exactly this in https://github.com/typelevel/scalacheck/pull/838#discussion_r931864166 but I obviously didn't understand 😂 sorry about that!

AurelienRichez: I think we were all confused by the name 😝. Good point about using `Prop.Exception`. It will make clear we're not talking about the usual `Exception`. I'll update it (it's not the only match on `Prop.Exception` in the file though).

I think scalac infer `{ case NonFatal(e) => e.getClass() }` as a `PartialFunction[Any, Class[_ <: Throwable]]` since that's what makes it compile, so it does not give a warning :-( . Intellij points to the error though. Unfortunately I was not using it in my first PR.

rossabaker (approved):

PR #841, opened 2021-10-06
Deterministically vary the number of seed slides in Test.check

leviramsey:
A `Prop.forAll` of multiple parameters will use the initial seed to generate the first parameter and slide the seed once for each of the subsequent parameters.  When combined with `Test.check` which slides the seed before evaluating the property, this means that for a property with _m_ > 1 parameters, if the _i_ th evaluation of the property had seeds (seed1 ... seedm), then the _i_+1 th evaluation will have seeds (seed2 ... seedm, seedm.slide).  If the same generator is used for multiple parameters, then a generated value in one evaluation will be reused in a different position in a specific later evaluation.

This addresses this by varying the number of slides done in `Test.check` based on the number of times a given property has been evaluated.  The number of slides remains deterministic, just less amenable to pattern recognition.

Issue #840, opened 2021-10-06
Predictable repetition in generated values


leviramsey:
See scalatest/scalatest#2054

```scala
trait Arithmetic {
  def add(x: Int, y: Int): Int
}

object ArithmeticProps {
  val arithmetic: Arithmetic = _ + _

  val dependsOnInput =
    Prop.forAll { (x: Int, y: Int, z: Int) =>
      println(s"$x $y $z")

      val addXY = arithmetic.add(x, y)
      val addXZ = arithmetic.add(x, z)

      (addXY == addXZ) == (y == z)
    }
}
```

`ArithmeticProps.dependsOnInput.check()` results in output like

```
0 1 -467377710
1 -467377710 -533146462
-467377710 -533146462 1
-533146462 1 1258003880
1 1258003880 1
1258003880 1 -830507445
1 -830507445 -1595422142
```

Note that the `x` value for a given run is the previous `y` and the `y` value for a given run is the previous `z`.

leviramsey:
This behavior was not observed in 1.14.3 (by way of the scalatest integration).

leviramsey:
This looks to be downstream of the seed handling change in #651 and will probably(?) only affect generating values where the generator only needs one random number (so `Int`/`Char`/etc.).

Varying the number of `reseed`s in `Test.check` (maybe `1 + (n + d) % 7` times?) should make this a little more coincidental.

ashawley:
ScalaCheck uses its own deterministic RNG.  It's meant for repeatability, but also tries to generate values quickly.  I predict what you're seeing in your example is an implementation detail of how the support for arity in`Prop.forAll` is done.  The multiple argument `Prop.forAll` is composed with calls to the `Prop.forAll` of one argument.  The one argument `Prop.forAll` calls `slideSeed` once, thus each parameter shares the same initial seed, but is only separated by one slide of the seed.  This explains why the same value is reproduced between arguments to `Prop.forAll`.

If I understand your addition example, this could seem wasteful since addition is commutative.  However, not all tests have that property.  I suppose the predictable repetition is odd compared to true RNG.  Even with parameter repetition, the tuples themselves are unique.

It looks like the case reported downstream with ScalaTest is related to generating two lists, but they're not distinct?

ashawley:
[...]

PR #911, opened 2022-08-11
Deterministically vary the number of seed slides in Test.check

leviramsey:
(re-attempt of #841, after that was inadvertently turned into an empty PR in the process of moving to a feature branch in my fork)

A `Prop.forAll` of multiple parameters will use the initial seed to generate the first parameter and slide the seed once for each of the subsequent parameters.  When combined with `Test.check` which slides the seed before evaluating the property, this means that for a property with _m_ > 1 parameters, if the _i_ th evaluation of the property had seeds (seed1 ... seedm), then the _i_+1 th evaluation will have seeds (seed2 ... seedm, seedm.slide).  If the same generator is used for multiple parameters, then a generated value in one evaluation will be reused in a different position in a specific later evaluation.

This addresses this by varying the number of slides done in `Test.check` based on the number of times a given property has been evaluated.  The number of slides remains deterministic, just less amenable to pattern recognition.

The seed sliding behavior between checks is configurable:

* If `maxRNGSpins` is `1`, `0`, or negative: 1 spin is performed
* Otherwise at most `maxRNGSpins` will be performed; the number of spins is deterministic and as the number of checks of a given property increases, the total number of spins performed between checks converges to `N * (maxRNGSpins + 1)/2`

### DIFF
--- a/core/jvm/src/test/scala/org/scalacheck/TestSpecification.scala
+++ b/core/jvm/src/test/scala/org/scalacheck/TestSpecification.scala
@@ -190,4 +190,28 @@ object TestSpecification extends Properties("Test") {
     val p1 = Prop(unique.size > 1) :| s"saw $n duplicate values: $unique"
     p0 && p1
   }
+
+  property("initialSeed is used and then updated when varying RNG spins") = {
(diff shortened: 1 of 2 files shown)


leviramsey:
Addresses #840 and scalatest/scalatest#2054

SethTisue:
Just one more thing before we merge — for the history, does this make sense as 6 commits?

PR #924, opened 2022-10-13
fix wrong millisecond value (fixes #913)

### LINKED ISSUE #913: Very minor issue in Gen.scala companion object
On lines 1519-1525, we see the following snippet:

```scala    
    val endOfDayGen: Gen[Calendar] =
      calendar.map(c => adjust(c) { c =>
        c.set(HOUR_OF_DAY, 23)
        c.set(MINUTE, 59)
        c.set(SECOND, 59)
        c.set(MILLISECOND, 59)
      })
```

The milliseconds are set to 59 when they should presumably be 999.

akonior:


rossabaker (approved):
Thanks!

PR #928, opened 2022-11-03
Fix Many Of The Warnings

isomarcte:
This is by no means all of them, but it is many of them, especially the very benign ones such as unused parameters.

There are a few more involved changes, such as deprecating a few methods and adding a few replacements. For example,

* The new `check_` methods in `Test`.
* The `with` methods in the `trait R` type.

### DIFF
--- a/core/shared/src/main/scala/org/scalacheck/Gen.scala
+++ b/core/shared/src/main/scala/org/scalacheck/Gen.scala
@@ -9,9 +9,6 @@
 
 package org.scalacheck
 
-import language.higherKinds
-import language.implicitConversions
-
 import rng.Seed
 import util.Buildable
 import util.SerializableCanBuildFroms._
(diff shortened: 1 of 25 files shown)


core/shared/src/main/scala/org/scalacheck/Prop.scala
```diff
@@ -43,7 +40,11 @@ sealed abstract class Prop extends Serializable { self =>
       res
     }
 
+  @deprecated(message="Please use useSeed which only takes a seed. name is, and has been, unused.", since="1.17.1")
```
rossabaker, line 43:
```suggestion
  @deprecated(message="Please use useSeed which only takes a seed. name is, and has been, unused.", since="1.18.0")
```

core/shared/src/main/scala/org/scalacheck/Gen.scala
```diff
@@ -410,28 +407,43 @@ object Gen extends GenArities with GenVersionSpecific {
 
     def retrieve: Option[T] = result
 
+    @deprecated(message="Please use withLabels, withResult, or withSeed instead.", since="1.17.1")
```
rossabaker, line 410:
```suggestion
    @deprecated(message="Please use withLabels, withResult, or withSeed instead.", since="1.18.0")
```

rossabaker (approved):
Thanks!  I'm updating the deprecation versions, because the next version will have to be a minor bump.

PR #943, opened 2023-02-10
Making the code more concise and readable

aashish2054:
The for-comprehension in the optimized code replaces the two exists statements, making the code more concise and readable. The code performs the same checks as the original code, but in a more elegant and readable way. Additionally, the yield statement evaluates to true if no initSut or runningSut matches the name of the new state, and false otherwise. The final line checks whether a runningSut matches the name of the new state, and returns the negation of this result. Overall, the optimized code is more concise, readable, and efficient than the original code.

### DIFF
--- a/examples/commands-leveldb/src/test/scala/CommandsLevelDB.scala
+++ b/examples/commands-leveldb/src/test/scala/CommandsLevelDB.scala
@@ -31,12 +31,6 @@ object LevelDBSpec extends Commands {
   }
 
   def canCreateNewSut(newState: State, initSuts: Traversable[State],
-    runningSuts: Traversable[Sut]
-  ) = {
-    !initSuts.exists(_.name == newState.name) &&
-    !runningSuts.exists(_.name == newState.name)
-  }
-  def canCreateNewSut(newState: State, initSuts: Traversable[State],
                       runningSuts: Traversable[Sut]
                      ) = {
     for {
--- a/examples/commands-leveldb/src/test/scala/CommandsLevelDB.scala
+++ b/examples/commands-leveldb/src/test/scala/CommandsLevelDB.scala
@@ -36,6 +36,19 @@ object LevelDBSpec extends Commands {
     !initSuts.exists(_.name == newState.name) &&
     !runningSuts.exists(_.name == newState.name)
   }
+  def canCreateNewSut(newState: State, initSuts: Traversable[State],
+                      runningSuts: Traversable[Sut]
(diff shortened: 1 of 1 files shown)


examples/commands-leveldb/src/test/scala/CommandsLevelDB.scala
```diff
@@ -31,9 +31,16 @@ object LevelDBSpec extends Commands {
...
+                      runningSuts: Traversable[Sut]
+                     ) = {
+    for {
+      initSut <- initSuts
+      runningSut <- runningSuts
+    } yield (initSut.name == newState.name) || (runningSut.name == newState.name) match {
+      case true => false
+      case false => true
+    }
+
```
rossabaker, line 43: Isn't this a discarded pure expression?  (I'm surprised scalac didn't complain?  Maybe I'm misreading it?)

rossabaker (commented):
Thanks.  I'm not sure how it's more concise, though?

examples/commands-leveldb/src/test/scala/CommandsLevelDB.scala
aashish2054: Can this be merged?

rossabaker: I might be missing something, but I still think that for comprehension is just being thrown away.  Does this code work as intended?

aashish2054: yeah it works

rossabaker (changes requested):
The tests are only compiled, not run, in CI.  I can't run them locally because my system does not have easy access to leveldbjni-1.8.  But the behavior definitely changes:

```scala
  val states = List(State(true, "issue-943", Map.empty))
  val suts = List.empty[Sut]

  println(LevelDBSpec.canCreateNewSut(
    State(true, "issue-943", Map.empty),
    states,
    suts
  ))
```

This now prints `true`, where it used to print `false`.  Additionally:

* The for-comprehension builds an `Iterable[Boolean]`, when the method returns `Boolean`.  It compiles, but it doesn't influence the result.
* Instead of matching booleans to invert them, why not just yield the inverted value directly?
* The original implementation short-circuited on the first decisive result.  Now it's processing the entire input.
* It used to be run in time proportional to the sum of the lengths of `initSuts` and `runningSuts`.  Now it's proportional to the product.
* Does performance even matter here?

PR #944, opened 2023-02-13
Add Gen.nonEmptyStringOf

satorg:
I totally realize that there are a lot of other ways to achieve the same functionality with the existing generators,
but I think it is a quite basic feature and can come handy when someone (me, personally) just needs to "grab and go" a non-empty string. So I believe it would complement the `nonEmptyListOf` and `nonEmptyMap` generators pretty well.


rossabaker (approved):
I've worked around this multiple times.  Big :+1:.

vivekragunathan:
Curious to know if something like `nonEmptySomeOf` (that complements `someOf`) would make sense. I am not sure if such a method exists already.

rossabaker:
@vivekragunathan I'd open a separate issue to discuss that one, but it doesn't sound unreasonable...

PR #950, opened 2023-03-24
Add scalafmt to the repository

satorg:
This is an effort to introduce ScalaCheck and ScalaFmt to each other.
Recently I realized that those two great guys have never ever met.
So I decided to give it a try.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,3 +1,7 @@
-# Header whitespace cleanup
+# Header organizing (PR: #959, #963)
 b8564609df3c849b3a258a411e0de645d4e8c7a0
 0b26b5e6dbcf3e6beeca806297f848f41dde899b
+
+# Code formatting (PR: #950)
+ccd98e22a078e1282dbb2c1acfb40fe3afee01b6
+
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -30,5 +30,3 @@ fileOverride {
 project.excludeFilters = [
   "scalafix/*"
 ]
-
-project.includePaths = [] # temporarily disable formatting
--- a/bench/src/main/scala/org/scalacheck/bench/GenBench.scala
+++ b/bench/src/main/scala/org/scalacheck/bench/GenBench.scala
@@ -16,44 +16,31 @@ import org.scalacheck.Arbitrary.arbitrary
 import org.scalacheck.Gen
 import org.scalacheck.rng.Seed
 
-/**
- * Generator benchmarks
- *
- * Since generators may run very quickly (or very slowly) depending on
- * the seed and size parameter used, we want to make sure these are
- * held constant across runs. Otherwise, we might believe a particular
- * change made a given generator faster (or slower) when in fact we
- * just got lucky (or unlucky).
- *
- * We use `seedCount` to choose how many seeds we will benchmark with.
- * For each seed, we run the given generator. So if `seedCount` is
- * 100, the average time we get is the time to generate values for the
- * 100 seeds we started with.
- *
- * The size parameter also plays a major role in how long generators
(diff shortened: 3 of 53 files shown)


satorg:
I've created as a Draft because before starting applying the formatting, I would like to agree on the scalafmt.conf file itself.
The initial scalafmt.conf is very minimalistic, but I am open to tune it as necessary to make everyone comfortable with it.

vivekragunathan:
It would be nice if we agree to apply the formatting to `build.sbt` too. 

Also I think we should takes a few rounds of applying the formatting to reduce change history overload. Start with something small and iteratively apply other (bigger) things.

satorg:
> It would be nice if we agree to apply the formatting to build.sbt too.

Yes sure – this PR implies formatting for build.sbt too.

rossabaker:
Don't forget that format commits can be added to a .git-blame-ignore-revs file.  Most Typelevel repos use the same config, as a lot of the benefit is to avoid fights over inconsequential preferences.  I'm in favor of this, and I think one isolated commit to do the reformat, added to blame ignore, would be great.

vivekragunathan:
> Don't forget that format commits can be added to a .git-blame-ignore-revs file. Most Typelevel repos use the same config, as a lot of the benefit is to avoid fights over inconsequential preferences. I'm in favor of this, and I think one isolated commit to do the reformat, added to blame ignore, would be great.

Oh wow, I did not that. That would work out great.

rossabaker:
A really cool thing is that Scala Steward adds to that file whenever the formatting changes through plugin upgrades.

project/plugin.sbt
```diff
@@ -3,6 +3,7 @@ addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.4.10")
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.2.0")
 addSbtPlugin("org.portable-scala" % "sbt-scala-native-crossproject" % "1.2.0")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.4.4")
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.0")
 val sbtTypelevelVersion = "0.4.18"
 addSbtPlugin("org.typelevel" % "sbt-typelevel-ci-release" % sbtTypelevelVersion)
 addSbtPlugin("org.typelevel" % "sbt-typelevel-settings" % sbtTypelevelVersion)
```
armanbilge, line 9: Now that we have a scalafmt config, if we're willing to throw in copyright headers than we can upgrade to `sbt-typelevel` proper:
```suggestion
addSbtPlugin("org.typelevel" % "sbt-typelevel" % "0.4.18")
```

rossabaker:
Also needs a `githubWorkflowGenerate`?

satorg:
@rossabaker , I'm struggling with the `startYear` setting, which is required for sbt-header.
Currently in the headers it is set in this way:
```
/*-------------------------------------------------------------------------*\
**  ScalaCheck                                                             **
**  Copyright (c) 2007-2021 Rickard Nilsson. All rights reserved.          **
...
```
But if I set it to 2007 and keep using the "BSD-3-Clause" license, then the headers become
```
/*
 * Copyright (c) 2007, Typelevel
 * All rights reserved.
...
```
Would such a change be fine there?

satorg:
I think this PR is ready for review.

Feel free to suggest changes to the `.scalafmt.conf` file please – I tried to be conservative in this initial work, but open to any other formatting rules.

Also, please pay attention to scaladoc comments – turned out they've changed the most after the re-formatting.

The last commit in the PR contains changes to `.git-blame-ignore-revs`.

Thank you.

rossabaker (approved):
This is good!  Sorry, it has been a chaotic week at $WORK and $HOME.

PR #959, opened 2023-05-30
Integrate sbt-header (via sbt-typelevel)

satorg:
This PR should technically precede #950.
Initially my intent was to just introduce scalafmt in here, but then I realized that it would be easier if we would have the **sbt-header** incorporated first. So here this PR is.

A couple of things to mention:
1. Most of the source files had their own headers already which do not correspond to any of the standard headers that **sbt-header** provides out of the box. So I created a custom header which is just a copy of the original one, but uses an **sbt-header** default style. See `project/CustomHeaderPlugin.scala`.
2. The **sbt-header** does not cover source files that are loaded into and used by SBT directly (i.e. those located in the `project` dir). However, **scalacheck** has `project/codegen.scala` along with the aforementioned `CustomHeaderPlugin.scala`. So I plugged them in to the sbt-header in `build.sbt` (see settings for the `root` project).
3. For the full version of **sbt-typelevel** the **.scalafmt.conf** should exist. So I created one but disabled any formatting in it for now. I think, it'd still be better to enable it in a separate follow up PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,10 @@ jobs:
         project: [rootJS, rootJVM, rootNative]
         workers: [1, 4]
         exclude:
+          - scala: 3.2.2
+            java: temurin@11
+          - scala: 2.12.17
+            java: temurin@11
           - project: rootJS
             java: temurin@11
           - project: rootNative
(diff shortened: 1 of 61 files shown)


rossabaker:
The approach seems fine to me, but if it simplifies things, I think the important factors is that Rickard maintains the copyright and the BSD3 license is preserved.  If there's a built-in config that satisfies that, it seems alright to me.

satorg:
> The approach seems fine to me, but if it simplifies things, I think the important factors is that Rickard maintains the copyright and the BSD3 license is preserved. If there's a built-in config that satisfies that, it seems alright to me.

There's BSD3 build-in header for sure, but it does not have any mention regards to Rickard.

My reasoning is that the current approach I'm suggesting does not makes things worse – i.e. it incorporates sbt-header, but still keeps the headers the same way. So that we can work out the exact headers content in one of the follow up steps.

So basically I'm just trying to eat that mammoth up in parts rather than trying to swallow the whole thing at once :)

rossabaker (approved):

satorg:
Seems good to go now.

PR #963, opened 2023-06-04
Fix trailing whitespaces in headers

satorg:
A follow up for #959.
When I resumed my work on #950, I realized that I overlooked trailing whitespaces in the header template in my last PR and Scalafmt started removing them.
It makes sbt-header and scalafmt incompatible with each other.
I decided to fix the header template and re-apply sbt-header.
Hopefully, it is the last PR regarding the headers and the next one is expected to be the Scalafmt per se.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Header whitespace cleanup
+0b26b5e6dbcf3e6beeca806297f848f41dde899b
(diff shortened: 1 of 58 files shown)


satorg:
@rossabaker could you take a look please? Thank you!

rossabaker (approved):

satorg:
@rossabaker thank you for adding .git-blame-ignore-revs!
Should I add this commit to that file as well:
[b8564609df3c849b3a258a411e0de645d4e8c7a0](https://github.com/typelevel/scalacheck/commit/b8564609df3c849b3a258a411e0de645d4e8c7a0)
?

rossabaker (approved):
@satorg, I think you have merge rights and I'm about to be in meetings all day.  Feel free to merge if it's green and you see it before me.

PR #991, opened 2023-07-27
Add `open` to `Properties` to prevent Scala 3 compile errors

froth:
When extending `Properties` in a Scala 3 code base one currently has to use `scala.language.adhocExtensions`. 
According to the documentation `Properties` is clearly intended for extension.


satorg (approved):
Makes sense to me. Thank you!

PR #1005, opened 2023-09-09
add explicit type

xuwei-k:
prepare Scala 2.13.12

- https://github.com/scala/scala/pull/10439

### DIFF
--- a/core/shared/src/main/scala/org/scalacheck/ScalaCheckFramework.scala
+++ b/core/shared/src/main/scala/org/scalacheck/ScalaCheckFramework.scala
@@ -155,7 +155,7 @@ private abstract class ScalaCheckRunner extends Runner {
             case _ => new OptionalThrowable()
           }
           def fullyQualifiedName() = self.taskDef().fullyQualifiedName()
-          def selector() = new TestSelector(name)
+          def selector(): Selector = new TestSelector(name)
           def fingerprint() = self.taskDef().fingerprint()
           def duration() = -1L
         }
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/org/scalacheck/Test.scala
```diff
@@ -369,7 +369,7 @@ object Test {
       val help = "Maximum number of RNG spins to perform between checks"
     }
 
-    val opts = Set[Opt[_]](
+    val opts: collection.Set[Opt[_]] = Set[Opt[_]](
```
satorg, line 372: Just wondering – can the type be `collection.immutabe.Set` rather than `collection.Set`? Or does it break compatibility?

core/shared/src/main/scala/org/scalacheck/ScalaCheckFramework.scala
```diff
@@ -155,7 +155,7 @@ private abstract class ScalaCheckRunner extends Runner {
             case _ => new OptionalThrowable()
           }
           def fullyQualifiedName() = self.taskDef().fullyQualifiedName()
-          def selector() = new TestSelector(name)
+          def selector(): Selector = new TestSelector(name)
           def fingerprint() = self.taskDef().fingerprint()
           def duration() = -1L
```
satorg, line 160: Out of curiosity – how come `def selector()` requires explicit type, but its neighbors (`fullyQualifiedName`, `fingerprint`, etc) do not?

satorg (approved):
Thank you for the PR!

satorg:
I think I can go ahead and merge it. Thank you @xuwei-k !

PR #1041, opened 2024-03-16
Enable lazy construction of label strings

som-snytt:
A different take on https://github.com/typelevel/scalacheck/pull/979

Instead of recycling `|:` for labeling, use `=|=` for improved symmetry and operator precedence.

A second commit follows up scalafmt config by "fixing" star alignment in doc comments. Edit: That hassle is reserved for another effort.

### DIFF
--- a/core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -583,6 +583,8 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
       }
       val avg = sum / n
       s"average = $avg" |: avg >= 0.49 && avg <= 0.51
+      s"average = $avg".ensuring(false, "eager evaluation") =|= avg >= 0.49 && avg <= 0.51
+      avg >= 0.49 && avg <= 0.51 =|= s"average = $avg"
     }
 
   property("uniform long #209") = {
--- a/core/shared/src/main/scala-2/org/scalacheck/package.scala
+++ b/core/shared/src/main/scala-2/org/scalacheck/package.scala
@@ -1,0 +1,29 @@
+/*
+ * ScalaCheck
+ * Copyright (c) 2007-2021 Rickard Nilsson. All rights reserved.
+ * http://www.scalacheck.org
+ *
+ * This software is released under the terms of the Revised BSD License.
+ * There is NO WARRANTY. See the file LICENSE for the full text.
+ */
+
+package org.scalacheck
+
+object `package` {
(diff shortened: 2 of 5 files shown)


som-snytt:
TIL scalafmt doesn't do headers.

core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
```diff
@@ -583,6 +583,8 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
       }
       val avg = sum / n
       s"average = $avg" |: avg >= 0.49 && avg <= 0.51
+      s"average = $avg".ensuring(false, "eager evaluation") =|= avg >= 0.49 && avg <= 0.51
+      avg >= 0.49 && avg <= 0.51 =|= s"average = $avg"
```
som-snytt, line 587: just to exercise the idioms in the absence of tests

som-snytt:
On precedence, the trade-off is the `&&` combinator for `Prop` (as opposed to the boolean op in an expression promoted to `Prop`).

core/shared/src/main/scala-2/org/scalacheck/package.scala
```diff
@@ -0,0 +1,29 @@
...
+
+object `package` {
+  implicit class `by-name label :|`(private val prop: Prop) extends AnyVal {
+
+    /** Label this property.
+      *
+      * The label is evaluated lazily. The operator name is chosen for its precedence btween boolean operators and
+      * others.
+      */
+    def =|=(label: => String): Prop = prop.map(_.label(label))
```
mrdziuban, line 20: Another part of my PR (#979) was to avoid evaluating the label if it will never be displayed to the user. What do you think about doing the same here?

```scala
def =|=(label: => String): Prop =
  prop.map(r =>
    r.status match {
      case Prop.False | Prop.Exception(_) => r.label(l)
      case Prop.Proof | Prop.True | Prop.Undecided => r
    })
```

core/shared/src/main/scala-3/org/scalacheck/package.scala
```diff
@@ -0,0 +1,21 @@
...
+
+package org.scalacheck
+
+object `package` {
+
+  /** Label this property. The label is evaluated lazily.
+    *
+    * The operator name is chosen for its precedence between boolean operators and others.
+    */
+  extension (prop: Prop) def =|=(label: => String): Prop = prop.map(_.label(label))
```
mrdziuban, line 18: Does Scala 3 not need an `extension` for `Boolean` like the Scala 2 version?

mrdziuban:
@satorg @som-snytt do you have any thoughts on moving this or #979 forward?

PR #1042, opened 2024-03-16
Prefer docstrings.style = AsteriskSpace

som-snytt:
Prefer the Scaladoc comment format used by Scala 3.

The two-space "gutter" makes it easier to read the source text. (Users of IDEs and published API will be indifferent to the source.)

It doesn't appear that the matter received any discussion last summer.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -11,6 +11,8 @@ danglingParentheses {
   callSite = false
 }
 
+docstrings.style = AsteriskSpace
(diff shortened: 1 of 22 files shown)


satorg:
I personally don't care – instead I usually prefer using as many default settings as possible. But if anyone feels more comfortable with some specific settings, I am for it.

satorg (approved):
Thank you!

PR #1044, opened 2024-03-25
Remove CODE_OF_CONDUCT, use org default [ci skip]

valencik:
As per https://github.com/typelevel/governance/issues/133, we're removing the individual CODE_OF_CONDUCT.md files so that org projects use the GitHub Typelevel organization default, which is easier for us to maintain.


satorg (approved):
Thanks!

PR #1062, opened 2024-07-14
Introduce `Gen.zipWith` multi-arity method

satorg:
Adds a code generator for `Gen.zipWith` functions.

Example:
```scala
case class Foo(pos: Int, ascii: String, hex: String)

Gen.zipWith(Gen.posNum[Int], Gen.asciiStr, Gen.hexStr)(Foo.apply)
```
Without `Gen.zipWith` the same was possible to accomplish with `Gen.zip`, however it would required an additional `map` over an intermediary tuple instance:
```scala
Gen.zip(Gen.posNum[Int], Gen.asciiStr, Gen.hexStr)
  .map { case (pos, ascii, hex) => Foo(pos, ascii, hex)) }
```
Therefore `Gen.zipWith` allows to avoid the intermediate conversion which can come in handy in come cases.

Note that `Gen.zipWith` starts with arity 2 whereas `Gen.zip` starts with arity 1.
However, I think the latter was an oversight – `Gen.zip` for arity 1 is a no-op function.

### DIFF
--- a/core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -511,7 +511,7 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
     // format: off
     val ts = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22)
     def fun(
-      t1: Int, t2 : Int, t3: Int, t4: Int, t5: Int, t6: Int, t7: Int, t8: Int, t9: Int, t10: Int, t11: Int,
+      t1: Int, t2: Int, t3: Int, t4: Int, t5: Int, t6: Int, t7: Int, t8: Int, t9: Int, t10: Int, t11: Int,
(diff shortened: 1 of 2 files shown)


rossabaker:
Is this intended to behave similarly to mapN, but without the Cats dependency?

satorg:
> Is this intended to behave similarly to mapN, but without the Cats dependency?

Pretty much. To be more accurate, it is supposed to resemble functions `map2` through `map22` of the `Apply` typeclass, because `mapN` requires a tuple to be created first, check this out:
```scala
case class Foo(pos: Int, ascii: String, hex: String)

val gen1 = (Gen.posNum[Int], Gen.asciiStr, Gen.hexStr).mapN { Foo.apply }

val gen2 = Apply[Gen].map3(Gen.posNum[Int], Gen.asciiStr, Gen.hexStr) { Foo.apply }

val gen3 = Gen.zipWith(Gen.posNum[Int], Gen.asciiStr, Gen.hexStr) { Foo.apply }
```

However, there's one catch with the Cats functions: they both use a chain of `Functor.product` under the hood, which in turn calls to `Gen.zip` with arity 2. I.e. the Cats functions incur a bunch of intermediary tuples on every evaluation to be created.

rossabaker (approved):
I lost track of this one in a GitHub bankruptcy.  I'm sorry it got so desperately lonely you had to talk to a bot.

PR #1070, opened 2024-09-01
Update `.scalafmt.conf`. enforce new syntax

xuwei-k:


satorg (approved):
Makes sense to me, thanks!

PR #1107, opened 2025-04-04
Increase Fidelity of the sbt.testing.Framework Implementation

### LINKED ISSUE #1105: sbt ScalaCheckRunner: loss of test selection fidelity
The only code that looks at the test selectors supplied via the `sbt.testing.TaskDef` is `ScalaCheckRunner.checkPropTask()`; it [recognizes](https://github.com/typelevel/scalacheck/blob/main/core/shared/src/main/scala/org/scalacheck/ScalaCheckFramework.scala#L116) `sbt.testing.TestSelector`s and selects the properties with the exact names supplied; it should _also_ recognize `sbt.testing.TestWildcardSelector`s and select the properties with the names containing the wildcard supplied.

Also, [ScalaCheckRunner.deserializeTask()](https://github.com/typelevel/scalacheck/blob/main/core/shared/src/main/scala/org/scalacheck/ScalaCheckFramework.scala#L29) should probably count `sbt.testing.TestWildcardSelector`s towards `countTestSelectors` and thus dispatch to `ScalaCheckRunner.checkPropTask()` and not to `ScalaCheckRunner.rootTask()` if any are supplied.

Finally, whatever test selection functionality `ScalaCheck` provides only kicks in in very specific use cases:
- when `ScalaCheckRunner.deserializeTask()` is called or
- when class name of the `sbt.testing.Fingerprint` used contains `"ForkMain"`.

When running `ScalaCheck` the "official" `sbt.testing` way, by calling `ScalaCheckRunner.tasks()` with the fingerprints returned from `ScalaCheckFramework.fingerprints()` (none of which contain `"ForkMain"` in their class names), `ScalaCheckRunner.checkPropTask()` is [never called](https://github.com/typelevel/scalacheck/blob/main/core/shared/src/main/scala/org/scalacheck/ScalaCheckFramework.scala#L44), and no test selection is applied.

Context: [Scala.js Gradle plugin](https://github.com/dubinsky/scalajs-gradle) integrates with `sbt.testing`-enabled test frameworks, and `ScalaCheck` is among them. Plugin does its own `TaskDef` serialization/deserialization in a way that is not specific to the test framework used, but works with Gradle. As a result of the above idiosyncrasies in the `ScalaCheck`'s test selection functionality, plugin can not access any of it :(

dubinsky:
- when `Selector`s supplied include only `TestSelector`s and `TestWildcardSelector`s, filter properties to run by matching their names against the `Selector`s;
- match a property by both its short and full name;
- added a test that actually runs ScalaCheck via the `SBT Test Interface` and demonstrates the (now correct) treatment of the `Selector`s;
- test also demonstrates two unfixable infidelities in the treatment of the nested properties;
- thankfully, ScalaCheck's implementation of `sbt.testing.Framework` is, unlike in some other test frameworks, shared between the platforms (JVM, Scala.js, Scala Native), so the fixes do not have to be replicated for each platform, but:
- the test needs to supply a `testClassLoader: ClassLoader` parameter when calling `sbt.testing.Framework.runner()`; on platforms other than JVM, `getClass.getClassLoader` is not available, so `Platform.getClassLoader: ClassLoader` method was added to every `Platform`; on platforms other than JVM, it returns `null` - which is fine, since on those platforms `sbt.testing.Framework.runner()` ignores the `testClassLoader` parameter anyway.

fixes #1105


dubinsky:
@satorg any thoughts?

dubinsky:
@satorg I am starting to suspect that the plan here is to let this fix wither on the wine like the previous contribution in this area (https://github.com/typelevel/scalacheck/pull/1031); if this is true, and my contribution is for some reason not wanted, please say so; otherwise - what do I need to do to get this merged?

Thank you!

dubinsky:
> Personally, I'd love to get both PRs (this and #1031) merged ultimately. I just don't feel very confident about this ScalaCheckFramework implementation, because I personally don't use it and mostly rely on other test framework runners.

Code changes have comments, tests are added that verify the change, and issue #1105 has detailed explanations. Is there anything else I can do? Are you going to involve other contributors/maintainers that are more familiar with this area or do you have specific questions?

> Do you think your PR complements #1031 or supersedes it completely?

My pull request completely supersedes #1031: it handles both `TestSelector` and `TestWildcardSelector`, whereas #1031 handles just `TestSelector`.

Thank you!

core/shared/src/main/scala/org/scalacheck/ScalaCheckFramework.scala
```diff
@@ -92,8 +92,23 @@ private abstract class ScalaCheckRunner extends Runner {
...
+          selector.isInstanceOf[TestWildcardSelector])
+
+      def isIncluded(name: String): Boolean = !isTestsOnly || td.selectors().exists {
+        case s: TestWildcardSelector => name.contains(s.testWildcard())
+        // For TestSelector, exact comparison with both the full name and the name with
+        // the suite prefix (properties.name) stripped off works for non-nested suite,
+        // but results in a false negative for nested suites: test "A.B.test" can not be
+        // selected by its short name "test";
+        // instead, selected test name is matched using `endsWith`.
+        case s: TestSelector => name.endsWith(s.testName())
```
satorg, line 107: @dubinsky , just to clarify: for nested test suites, shouldn't `NestedTestSelector` be used instead?

dubinsky: As far as I understand, for _selecting_ the tests to run, only `TestSelector` and `TestWildcardSelector` are used; `NestedTestSelector` should be used  for _reporting_ status of a test that is nested.

satorg:
@dubinsky , thank you for your work on this issue. I found some time to wrap my head around SBT testing API, so that now I think understand what is going on there to some extent. If you are still interested, I guess we can get your PR merged.

However, I wouldn't like to dismiss the effort made in #1031 too, since it was kinda preceding yours. How about if I merge the former PR first, then you'll update yours overwriting conflicting parts and we'll merge your PR right after? Thereby everyone will get involved. @Duhemm, in a case you are still around, let us know if you are OK with this plan please. Thanks everyone!

satorg:
@dubinsky , do you have any clue what `serializeTask`/`deserializeTask` methods of `ScalaCheckRunner` could be for? They look like they are supposed to be called via reflection, but I didn't manage to figure out when and how exactly.

dubinsky:
> @dubinsky , do you have any clue what `serializeTask`/`deserializeTask` methods of `ScalaCheckRunner` could be for? They look like they are supposed to be called via reflection, but I didn't manage to figure out when and how exactly.

They are called by the Test Interface/Test Bridge on [Scala.js](https://github.com/scala-js/scala-js/blob/main/test-bridge/src/main/scala/org/scalajs/testing/bridge/TaskInfoBuilder.scala#L19) and [Scala Native](https://github.com/scala-native/scala-native/blob/main/test-interface/src/main/scala/scala/scalanative/testinterface/TaskInfoBuilder.scala#L8) to transport the task information from JVM to Node.js/Native code.

satorg:
Hi @dubinsky , I've merged #1031.
Feel free to update this one and let's get it merged too.
Thanks!

dubinsky:
> update this one and let's get it merged too. Thanks!

Done. Thank you!

PR #1119, opened 2025-08-08
Fix Gen.pick producing only a subset of possible combinations

sirmax:
When calculating a random insertion index for an element `Gen.pick` uses wrong order of operations
```
val i = (x & Long.MaxValue % count).toInt
```
As written, it first gets the modulo `MaxValue & count` and then uses it to mask a random `x` to get a positive value. Instead it should mask the `x` first, and only then get a modulo of it.

This PR fixes the order of operations.

### DIFF
--- a/core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
+++ b/core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
@@ -293,12 +293,18 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
 
   property("pick produces enough distinct combinations") = {
     val (n, nElements, nCombinations) = (3, 5, 10)
+    // The number of samples to produce with Gen.pick, that is enough to assert
+    // that every possible combination appears at least once. It depends on
+    // nSamples, but there is no specific math behind it. Rather, it's just an
+    // empirically chosen value, that has yielded no failures over thousands of
+    // test cycles.
+    val nSamples = 200
 
     val genPick = pick(n, 1 to nElements)
       // not interested in different permutations, only in distinct combinations
       .map(_.sorted)
 
-    forAllNoShrink(listOfN(100, genPick).map(_.distinct)) { distinctCombinations =>
+    forAllNoShrink(listOfN(nSamples, genPick).map(_.distinct)) { distinctCombinations =>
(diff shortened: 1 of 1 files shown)


core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
```diff
@@ -291,6 +291,18 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
     }
   }
 
+  property("pick produces enough distinct combinations") = {
+    val (n, nElements, nCombinations) = (3, 5, 10)
+
+    val genPick = pick(n, 1 to nElements)
+      // not interested in different permutations, only in distinct combinations
+      .map(_.toList.sorted)
```
satorg, line 299: Out of curiosity, what is the reason for converting the output of `pick` to `List` before sorting it? It doesn't seem it is supposed to improve performance or affects the test in any other way.

sirmax: It looks like a leftover from my attempts at writing this test. Removed it.

I think, at some point I had sets and also tried to collect the stats on those. It went into `.toString` on sets which, amusingly, may be different for equal small sets (up to 4 elements). Thus, it ended up with `aSet.toList.sorted` and later got scrapped, but `toList` somehow survived 😅

core/jvm/src/test/scala/org/scalacheck/GenSpecification.scala
```diff
@@ -291,6 +291,24 @@ object GenSpecification extends Properties("Gen") with GenSpecificationVersionSp
     }
   }
 
+  property("pick produces enough distinct combinations") = {
+    val (n, nElements, nCombinations) = (3, 5, 10)
+    // The number of samples to produce with Gen.pick, that is enough to assert
+    // that every possible combination appears at least once. It depends on
+    // nSamples, but there is no specific math behind it. Rather, it's just an
+    // empirically chosen value, that has yielded no failures over thousands of
+    // test cycles.
```
sirmax, line 300: > …over thousands of test cycles.

Specifically this
```
+ Gen.pick produces enough distinct combinations: OK, passed 10000000 tests
```

satorg (approved):

PR #1167, opened 2026-04-11
Added live site with typelevel site and laika.

dhinojosa:
Note:
The Old Releases Page may not work as we move away from the old scalacheck.org website. The old website has a `files/` folder that still needs to be up and running for that page to work. We may no longer want to maintain the very old releases, so we can stop advertising them.


docs/files/index.html
```diff
@@ -0,0 +1,921 @@
+<!DOCTYPE html>
```
satorg, line 1: It looks like a pre-rendered HTML in here.
Shouldn't it be in Laika sources in Markdown format insead?

dhinojosa: So that's the slide deck from Rickard. See the current website here. https://scalacheck.org/files/scaladays2014/index.html#1

I'm bringing it in so it can still be seen as a slide deck. To see how it is generated, see the branch `website` and the directory _src/articles/scaladays2014/presentation/_. The original deck is actually in markdown, but the slide deck is rendered with a _default.nix_ file, in the same _presentation_ folder that uses a remark.js derivation. 

I can either:

1. Keep it as an HTML/CSS already built static file the way it is now, since I don't think any more changes are going to be made
2. Just post the slide deck as markdown - it won't look and feel like a slide deck anymore, but just like another page on the website.
3. Try to make it part of the build somehow, and build it with remark.js and link it somehow to the project
4. Just remove the slide deck entirely as a resource and just leave the link to the parlays talk on YouTube.
5. Your ideas here

dhinojosa:
FYI, I changed the _old_releases.md_ page so it no longer relies on the old website.

docs/files/index.html
satorg: @dhinojosa ,
> 2. Just post the slide deck as markdown - it won't look and feel like a slide deck anymore, but just like another page on the website.

I'd probably go with that option. A regular markdown should be easier to read and easier to search its content up.

[link](https://github.com/typelevel/scalacheck/blob/website/src/articles/scaladays2014/presentation/presentation.markdown#the-commands-api-needs-more-evaluation-and-polishing-will-land-in-scalacheck-1120)

@armanbilge , wdyt?

dhinojosa: Let me know and I will tidy it up @armanbilge @satorg

docs/files/remark.js
armanbilge, line 1: We can load this over a CDN instead of committing to the repo. Not sure what version it is.

https://cdnjs.com/libraries/remark

build.sbt
```diff
@@ -69,6 +69,114 @@ ThisBuild / tlMimaPreviousVersions ++= Set(
...
+import laika.config.{LinkConfig, MessageFilters, SyntaxHighlighting, TargetDefinition, Version, Versions}
+import laika.helium.Helium
+import laika.helium.config.{Favicon, HeliumIcon, IconLink, LinkPanel, ReleaseInfo, Teaser, TextLink}
+import laika.ast.Path.Root
+import laika.ast.Image
+import laika.format.Markdown
+import laika.theme.config.Color
+
+val scalaCheckRepo = "https://github.com/typelevel/scalacheck"
+val latestVersion = "1.19.0"
```
armanbilge: Instead of hard-coding these, we should read them out of settings. For example, the repository is available as `scmInfo.value.map(_.browseUrl)`.

build.sbt
```diff
@@ -69,6 +69,114 @@ ThisBuild / tlMimaPreviousVersions ++= Set(
...
+
+lazy val docs = project.in(file("site"))
+  .enablePlugins(TypelevelSitePlugin)
+  .settings(
+    laikaExtensions ++= Seq(
+      Markdown.GitHubFlavor,
+      SyntaxHighlighting
+    ),
```
armanbilge, line 89: I think the `TypelevelSitePlugin` already enables these things.

docs/sources.md
armanbilge, line 1: I don't think this page is very meaningful anymore 😅

docs/files/index.html
armanbilge: I'm okay to keep this as html. Let's try to preserve the original URL.

armanbilge (commented):
thanks for your work on this!

docs/files/remark.js
dhinojosa: Finally, getting back to all this. Sorry, having too much fun. 

The Nix files that build the static site are in the other branch:
 https://github.com/typelevel/scalacheck/blob/website/src/articles/scaladays2014/presentation/default.nix

Where the _remark.js_ application is 
```nix
  src = fetchurl {
      url = "https://github.com/gnab/remark/archive/v0.13.0.tar.gz";
      sha256 = "0ds19h17ya1xvk8jyphdyc7zs4l90210qknldg0j9fb9n3q45ad8";
    };
```

docs/files/index.html
dhinojosa: > I'm okay to keep this as html. Let's try to preserve the original URL.

Question. Who has access to the CDN? Perhaps you, @armanbilge? And shall I assume the same URL, or wait for a new URL to post on the repo?

build.sbt
dhinojosa: If I take it out, it looks plain
image

If I leave the call in, it looks great
image

dhinojosa:
Updated with recommendations

satorg:
@dhinojosa ,
> Updated with recommendations

I missed your last update, sorry about that.

> This branch has conflicts that must be resolved

Could you resolve the conflicts please? I'm looking forward to have it merged!
Thank you!

